PR #177, opened 2014-06-16
SubFind HDF support

rokroskar:
- [x] create a SubFindHDFSnap class that extends GadgetHDFSnap and reads in the particle data -- this is mostly completed in 5f8710c but needs to be tested
- [x] create a SubFindHDFHaloCatalogue class that reads the FOF and subhalo group data and creates/returns the appropriate halos
- [x] add support for loading subhalos of fof groups, e.g. ``h[0].sub[0]`` (done in in 13bd72a)
- [x] loading for large data sets, where not all particle types might be present in all hdfs
- [x] fix unit tracking so that it doesn't require a full data load
- [x] get proper unit parsing for all of the subfind-derived properties
- [ ] allow efficient indexing of sub-group properties, e.g. ``h.sub_properties['MassType']`` should return all the masses for all subgroups, and ``h[1].sub_properties['MassType']`` should return all masses for subgroups in FOF group 1. 
- [x] documentation/tutorial 
- [x] make reading of large datasets possible -- HDF5 apparently dies when trying to open too many files, so replace the list expressions e.g. `self._hdf = [...]` with generators as a first try

test data here: https://t.co/MjfJdcmlLP
SubFind description here: https://t.co/znFJu7F9Xz


astroduff:
There are two outstanding issues my student raised (along with solutions that seem fine but I wanted to confirm before pushing)

Firstly when opening a number of subfiles for large runs (say the 512^3 sims with a similar number of subfiles) there appears
to be a crash as h5py doesn't close the subfiles after using. 
For instance, in around line 410 of gadgethdf.py:
 if not h5py.is_hdf5(filename) :                    
     numfiles = self.properties['NTask']            
     self._hdf = [h5py.File(filename+"."+str(       
         i)+".hdf5", "r") for i in xrange(numfiles)]

To solve this Yuxiang added a close function in the class, e.g. SubFindHDFSnap :
def close(self):          
    for item in self._hdf:
        item.close()      

So every time I use s = pn.load(FILE), after finish reading the data, I would close it by s.close()

Then secondly there is a failure when loading files with few subhalos (early subfind outputs) as halo.py line 1421:
ndim = len(sub_properties[key])/self.nsubhalos 

Here, len(sub_properties[key]) is an int number, which means dim*self.nsubhalos probably becomes smaller than len(sub_properties[key]). 
I can simply skip this error by putting sub_properties[key] = sub_properties[key].reshape(self.nsubhalos,ndim) into the try function. 

Not sure if this is a good idea or not?


The final outstanding issue is the slow time to access halo properties as a dataset, currently has to be an embedded
for loop over haloes and subhaloes, appending to a dataset... e.g. 
 In [17]: SubStellarVelDisp = []
 In [18]: for halo in h: 
 In [18]:       for subhalo in halo.sub:
 In [18]:               SubStellarVelDisp.append(subhalo.properties['SubStellarVelDisp'])

This is way beyond my familiarity with the code and would be great to resolve too but it's not needed for the merge I would say.

Cheers
Alan

rokroskar:
hey @astroduff can you make the test data available again? it seems to have disappeared from dropbox.

apontzen:
Hi all. Thanks for this. It seems like it's almost ready :-)

I am a bit confused about the `close` issue. What is the problem with the files remaining open, when should they be closed and why? Although adding a `close` method sounds like it gets around the problem, it adds a layer of (format-dependent) complexity that I'd like to avoid if possible. If I understand the problem better I may be able to see a way around it.  Perhaps the test data will help with this too.

I also don't quite get why the reshaping for the subhalos might fail. Surely if there are n subhalos there are always n*num_properties sub_properties, even if n is small? Then it seems like 1421 must always work to get num_properties. Or have I missed something here?

rokroskar:
The close problem seems to have been fixed -- replaced the list with a
generator.

astroduff:
To demonstrate the restructure issue I've loaded up example subhalo files 
that crash the code (53) :
https://copy.com/CYc447sw7dCNdKyQ

and the first output step that doesn't (54), can get:
https://copy.com/suHOaoJupVgZwkvk

Be warned they're pretty sizeable!

Cheers
Alan

astroduff:
I fixed the reshaping issue, as it was written some arrays were n_groups FOF groups long and some were n_subhalos Subhalos long.
By dividing the int by an int to check dimensionality it wasn't always catching all cases. I've now used % to search for exactly which 
array size is appropriate, and then calculate the dimensionality and reshape.

The close issue might be better solved in future but right now seems to be working well enough? 

Is it possible to merge the HDF version back into the main pynbody so I can tell collaborators to simply install pynbody from pypi 
rather than grabbing this branch on github?

Cheers
Alan

apontzen:
Yes in principle, though it looks like the merge is going to be dangerous and difficult! Also, do we have some unit tests for this work?

rokroskar:
yes the merge is going to be painful. I wonder if there is a way to do it
in stages? Commit partial merges? I think some of the things I fixed in the
general gadget HDF class have subsequently also been fixed by other people.

The unit tests are in nose/gadgethdf_test.py

apontzen:
@rokroskar if you are willing to do the work I would suggest creating a new branch off pynbody/master, merging carefully into that, then create a PR for that... I will review and do the actual final merge! This way we get two people casting their eye over it to check we're not accidentally reverting things (which seems to be a common problem with complex merges in the past).

rokroskar:
hmm ok I think that sounds reasonable. I guess all I need to do is to add
my fork as a remote repo to my local main pynbody repository, fetch the
subfind_hdf branch and merge in master. correct?

apontzen:
Yes I think that is right

astroduff:
Will I be able to gain access to this to try and help on the merge?
Or will the extra loop with this make things too complex??

Cheers
Alan

rokroskar:
@astroduff yes of course -- I just committed 4543441 which is a first attempt at the merge -- it seems to be failing unit tests for the moment so more work is required, but the merge itself was fairly painless so far

astroduff:
Great thanks man, I've made one last extensive change of the units for subfind 
files to ensure that it reads in the vast majority or arrays now and actually checks
if the conversion makes sense. 

Cheers
Alan

PR #203, opened 2014-09-05
Halo shape

stinsong4100:
This pull request adds a shape calculation based on the inertia tensor.  It might be able to optimized more.  At this stage, I left in debugging plots and timing code.  Delete these before merging!

Issues:
Slowness:  This method is iterative with a rotation every iteration, so recalculating 'r' is the slowest part even though 'r' is not used during the iteration.  Is there a way to turn off the 'r' re-calculation temporarily?

Does anyone know why the b and c axes flip?  In my test case, I'm getting a shorter b than c.


apontzen:
Superseded by another implementation

PR #236, opened 2015-02-05
Speedup sphere selection

rokroskar:
Stringing together a large number of filters is slow because the loop is over the full output for each filter. This PR modifies the `_sphere_selection` function to take an array of centers and radii and only loops over the full particle array once. 

It works something like this: 

```
s = pynbody.load(simulation)
filt = filt = pynbody.filt.Sphere('0.1 kpc', s.s['pos'][:2000])
```

this creates a filter that includes all particles within a radius 0.1 kpc around the first 2000 stars.

### DIFF
--- a/pynbody/filt.py
+++ b/pynbody/filt.py
@@ -139,7 +139,7 @@ class Sphere(Filter):
         if units.is_unit(self.radius[0]):
             return " & ".join(["Sphere('%s', %s)" % (str(radius), repr(cen)) for radius, cen in zip(self.radius, self.cen)])
         else:
-            return "Sphere(%.2e, %s)" % (self.radius, repr(self.cen))
+            return " & ".join(["Sphere(%.2e, %s)" % (radius, repr(cen)) for radius, cen in zip(self.radius, self.cen)])
 
 
 class Cuboid(Filter):
--- a/pynbody/filt.py
+++ b/pynbody/filt.py
@@ -109,8 +109,10 @@ class Sphere(Filter):
         if isinstance(radius[0], str):
             radius = map(units.Unit,radius)
 
-        if len(cen) != len(radius) : 
-            radius = radius * len(cen)
+        # check if many different centers were passed in
+            if isinstance(cen, list) or isinstance(cen, np.ndarray) : 
+                if len(cen) != len(radius) : 
+                    radius = radius * len(cen)
 
         self.radius = radius
 
@@ -124,6 +126,8 @@ class Sphere(Filter):
         if units.is_unit_like(radius[0]):
             radius = np.asarray([r.in_units(pos.units, **pos.conversion_context()) for r in radius], dtype='float')
 
+        else : 
+            radius = np.asarray(radius, dtype='float')
 
         cen = self.cen
         if units.has_units(cen):
--- a/pynbody/filt.py
+++ b/pynbody/filt.py
@@ -100,13 +100,20 @@ class Sphere(Filter):
     def __init__(self, radius, cen=(0, 0, 0)):
         self._descriptor = "sphere"
         self.cen = np.asarray(cen)
-        if self.cen.shape != (3,):
-            raise ValueError, "Centre must be length 3 array"
+#        if self.cen.shape != (3,):
+#            raise ValueError, "Centre must be length 3 array"
 
-        if isinstance(radius, str):
-            radius = units.Unit(radius)
+        if not isinstance(radius, list) : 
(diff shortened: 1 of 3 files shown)


pynbody/filt.py
```diff
@@ -100,36 +100,50 @@ class Sphere(Filter):
     def __init__(self, radius, cen=(0, 0, 0)):
         self._descriptor = "sphere"
         self.cen = np.asarray(cen)
-        if self.cen.shape != (3,):
```
apontzen: Removing this check could lead to a segfault in _sphere_selection

You could implement it as self.cen.shape[-1]!=3?

pynbody/filt.py
```diff
@@ -100,36 +100,50 @@ class Sphere(Filter):
...
-            raise ValueError, "Centre must be length 3 array"
+#        if self.cen.shape != (3,):
+#            raise ValueError, "Centre must be length 3 array"
 
-        if isinstance(radius, str):
-            radius = units.Unit(radius)
+        if not isinstance(radius, list) : 
+            radius = [radius]
```
apontzen: What if someone chooses to pass in radius as an ndarray?

pynbody/filt.py
```diff
@@ -100,36 +100,50 @@ class Sphere(Filter):
...
+
+        if isinstance(radius[0], str):
+            radius = map(units.Unit,radius)
+
+        # check if many different centers were passed in
+            if isinstance(cen, list) or isinstance(cen, np.ndarray) : 
+                if len(cen) != len(radius) : 
+                    radius = radius * len(cen)
```
apontzen: This line will have a different effect if radius is an ndarray

pynbody/filt.py
```diff
@@ -100,36 +100,50 @@ class Sphere(Filter):
...
 
     def __repr__(self):
-        if units.is_unit(self.radius):
-
-            return "Sphere('%s', %s)" % (str(self.radius), repr(self.cen))
+        if units.is_unit(self.radius[0]):
+            return " & ".join(["Sphere('%s', %s)" % (str(radius), repr(cen)) for radius, cen in zip(self.radius, self.cen)])
         else:
-            return "Sphere(%.2e, %s)" % (self.radius, repr(self.cen))
+            return " & ".join(["Sphere(%.2e, %s)" % (radius, repr(cen)) for radius, cen in zip(self.radius, self.cen)])
```
apontzen: Should this be | rather than &?

pynbody/filt.py
rokroskar: indeed, this needs to be made more general -- same with the comment below.

pynbody/filt.py
salomanders: I want to do a ~filter (cut out a sphere of particles to not include in the snapshot), and when it strings them together it only gives the ~ to the first sphere. 

filt = ~pynbody.filt.Sphere('0.1 kpc', st['pos'][0:4])
~Sphere('1.00e-01 kpc', array([  910.50829655,  1563.14880045,  1451.38020969])) & Sphere('1.00e-01 kpc', array([  913.86336456,  1566.48238759,  1451.69728888])) & Sphere('1.00e-01 kpc', array([  913.35059333,  1566.05790408,  1451.9326327 ])) & Sphere('1.00e-01 kpc', array([  909.26119045,  1565.22082582,  1449.37438321]))

rokroskar: Yes the notation is wrong -- the entire selection is negated. So the filter
originally *includes* all particles in the spheres but if you ~ it then
it's the inverse. It's an OR initially even though it says it's an AND...
Needs to be fixed

apontzen: Actually wouldn't it just be neatest to be honest that it's a single operation and have __repr__ return `Sphere(ndarray(...),ndarray(...))` rather than pretend it's a string of operations?

rokroskar: sure! I don't really care what it looks like, either way the string representation is a complete mess so maybe we could do something sensible like show the first couple and then have '...' at the end. The user can still access the radii and centers from the object itself should it be needed.

salomanders:
Getting back into this analysis, did this branch not get merged into master?

apontzen:
It looks like there are still a few fixes noted above that make it not ready - @rokroskar ?

rokroskar:
not sure when I'll be able to get to this -- the fixes look pretty minor

salomanders:
I am happy to take on the task, but not sure where to start or what the conflicts are

PR #240, opened 2015-02-24
Percentile profiles

jakobherpich:
For any simulation quantity `qty` (e.g. self defined values using `@pynbody.derived_array`), this pull request introduces the possibility to calculate the percentile of the respective quantity for every bin by invoking `prof['qyt_percentile50']`. This particular case returns the median. The general syntax is the quantity name followed by `"_percentile"` and any string representation of the percentage, e.g. '12.5', '16', '2e1' as long as the respective number is in the range 0...100.
```
prof = pynbody.analysis.profile.Profile(sim)
prof['vphi_percentile16.5']
```

### DIFF
--- a/pynbody/analysis/profile.py
+++ b/pynbody/analysis/profile.py
@@ -379,11 +379,16 @@ class Profile:
             return self._profiles[name]
             # else :
             #    raise RuntimeError, "Derivatives only possible for profiles of fixed bin width."
+        elif "_percentile" in name and name[:name.find("_percentile")] in self.keys() or name[:name.find("_percentile")] in self.derivable_keys() or name[:name.find("_percentile")] in self.sim.all_keys():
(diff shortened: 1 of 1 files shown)


apontzen:
This is a very nice facility... Could you add an example in one of the tutorials and/or a test? That would round it off nicely!

jakobherpich:
While adding an example to the tutorials, I realized that this feature has already been implemented in the `profile.QuantileProfile` class.

rokroskar:
yes this was added by @stinsong4100 a long time ago -- I guess documentation was never created for it. Would be useful to do so now @jakobherpich?

rokroskar:
hey @jakobherpich any chance to finish this off by adding documentation for the percentile profiles?

apontzen:
This appears to be abandoned; I'm closing the PR now.

Issue #261, opened 2015-08-02
Old config files cause new versions to fail importing


perretv:
I can use pynbody up to the commit 4938c7a5194c3da836cdf576310c9e0ab0b79f5a on a linux cluster, but after this commit if I try to import pynbody, I get the following error:

```
----> 1 import pynbody

/home/ics/vperret/python/default.env/plat/pynbody/__init__.py in <module>()
    176 
    177 # Import subpackages
--> 178 from . import util, filt, array, family, snapshot,  tipsy, gadget, gadgethdf, ramses, grafic, analysis, halo, derived, bridge, gravity, sph, nchilada, transformation
    179 
    180 try:

/home/ics/vperret/python/default.env/plat/pynbody/filt.py in <module>()
     15 
     16 import numpy as np
---> 17 from . import units
     18 
     19 class Filter(object) :

/home/ics/vperret/python/default.env/plat/pynbody/units.py in <module>()
    753 
    754 for new_unit_name, new_unit_definition in config_parser.items("named-units"):
--> 755     new_unit = NamedUnit(new_unit_name, new_unit_definition)
    756     if __is_clean_name(new_unit_name):
    757         globals()[new_unit_name] = new_unit

/home/ics/vperret/python/default.env/plat/pynbody/units.py in __init__(self, st, represents)
    350         self._st_rep = st
    351         if isinstance(represents, str):
--> 352             represents = Unit(represents)
    353 
    354         self._represents = represents

/home/ics/vperret/python/default.env/plat/pynbody/units.py in Unit(s)
    659                 u = _registry[s[0]]
    660             except KeyError:
--> 661                 raise ValueError("Unknown unit " + s[0])
    662             p = Fraction(s[1])
    663             if p.denominator is 1:

ValueError: Unknown unit rad
```

I use the following pydistutils.cfg:

```
[install]
install-base=$HOME/python/default.env
install-purelib=$base/site-packages
install-platlib=$base/plat
install-scripts=$base/scripts
install-headers=$base/include
install-data=$base/data
```
with ```PYTHONPATH=$HOME/python/default.env/plat```

Any idea on the origin of this error?

apontzen:
It seems like a problem with a configuration file. Can you post your .pynbodyrc and config.ini?

perretv:
I don't use any ```config.ini``` file, but here is my ```.pynbodyrc```:

```
# DO NOT alter this file directly
#
# If you want to override options, either create a local copy called
[...]

PR #262, opened 2015-08-03
Attempt sensible solution to issue #261 by merging the list of irreducible-units

apontzen:
…but issuing a warning when doing so.

@rokroskar could you take a look and see if this seems a good fix to you? It's a bit hacky...


apontzen:
Merged it anyway as I heard from someone else facing the same problem and it seems important to help people with this...

Issue #258, opened 2015-07-29
Issues request efficient subgroup properties with SubFindHDFSnap class


astroduff:
To speed up use of SubFind HDF5 group data for Smaug / OWLS using SubFindHDFSnap can we allow efficient indexing of sub-group properties, e.g. h.sub_properties['MassType'] should return all the masses for all subgroups, and h[1].sub_properties['MassType'] should return all masses for subgroups in FOF group 1.

rokroskar:
can you remind me, is this already calculated and stored somewhere in the hdf5 file? If I remember correctly, we were thinking about computing it when the halo catalogue is first being created, but I don't know how that is more efficient than what you have to do now which I guess is loop through all the subhalos?

astroduff:
So the pre-computed extra array properties from SubFind are available
which has helped mitigate things, for example previously to find the halo 
mass for, say, gas previously was:

import pynbody as pn
test = '~/NOSN_NOZCOOL_L010N0128/data/subhalos_103/subhalo_103'
s = pn.load(test)
h = s.halos()

h[0].g['Mass'].sum()

whereas now it's

h[0].properties['MassType'][0]

(although these are seemingly lacking units still which I can't quite get right).

However, if you want to define a temperature limited sample:

h[0].g['Mass'][h[0].g['temp'] < 4.5e4].sum()

Then you'd have to loop over each halo individually, or worse yet if interested
in subhaloes you'd do

cold_sub = np.zeros(s.properties['Total_Number_of_subgroups'])
counter = 0
for halo in h:
	for subhalo in halo.sub:
		cold_sub[counter] = subhalo.g[‘mass’][subhalo.g['temp'] < 4.5e4].sum()
		counter +=1

Not sure if there's any way around this though? 

Cheers
Alan

rokroskar:
I see -- I think for custom filters it will not be possible to create some sort of general aggregate functionality. I guess what you could do is to add a method to `SubfindHDFHaloCatalogue` and `SubfindHDFSubhaloCatalogue` with the boilerplate code above -- e.g.: 

```
def compute_aggregate(hc, array, family = None, filter = None) : 
    agg = 0
    for halo in hc : 
        if family is not None : 
            halo = halo[family]
        if filter is not None : 
            halo = halo[filter(halo)]
        agg += halo[array].sum()
```

actually, come to think of it, this should be a method of the `HaloCatalogue` class and work for all halo catalogues.

PR #269, opened 2015-08-09
Refactor GadgetHDFSnap and related classes, and add ability for flexible naming...

apontzen:
...so that reader works eg. for Gizmo files

This is a major refactorization to improve maintability and remove code duplication from GadgetHDFSnap and SubfindHDFSnap. It was triggered by trying to add the flexibility to load Gizmo files with the same reader (which is now possible). It involves breaking a lot of very long functions into smaller named functions to improve clarity, and removes a lot of duplicated code by using these new smaller functions.

@rokroskar and @astroduff it would be great if you can test this - although the unit tests pass, I am not convinced about whether they'd pick up on more subtle broken aspects.

I have still to properly refactor SubfindHDFCatalogue but this will have to wait for another day.

BTW, if you are wondering why I decided to clean this up, I highly recommend Clean Code by Bob Martin https://books.google.co.uk/books?id=dwSfGQAACAAJ&hl=en ;-)


astroduff:
So I've taken a look at that and one thing it broke was my (awful) 
created arrays such as 

h[0].g['HI']

Cheers
Alan

apontzen:
Thanks for testing Alan - but I can't reproduce the problem. Could you post the full failure case?

```python

In [1]: import pynbody

In [2]: f = pynbody.load("testdata/Test_NOSN_NOZCOOL_L010N0128/data/subhalos_103/subhalo_103")

In [3]: h = f.halos()

In [4]: h[0].gas['HI']
Out[4]: 
SimArray([  6.63111825e-06,   6.21908409e-06,   3.27587015e-06, ...,
         2.33669819e-08,   4.29318764e-08,   2.03932252e-08])
```

apontzen:
BTW you just made me realize that your ionfrac code was calling a .c fragment that is no longer part of pynbody. I guess it worked for both of us because we had installed over the top of old versions. I have updated it to use the correct .pyx module. Note that this gives (slightly) different results as the original .c interpolation routine was inaccurate.

astroduff:
Hi Andrew

Apologies for the silence, I've been away from the laptop
this last week. So everything looks fine in the latest version
(and yes thanks for letting me know about this modification
to the interpolation as it would have driven me crazy trying 
to figure out how the numbers had changed slightly otherwise)

Currently there appears to be no way to use pre-computed arrays 
such as:

h[0].sub[0].properties['Mass']

in a single access attempt (as it's a list) for example 

np.sum( h[0].sub[:].properties['Mass'] )

and instead access is via a for-loop

mass = 0
for subhalo in h[0].sub:
	mass += subhalo.properties['Mass']

Is there a way to change this properties to an array rather than 
list of SimArray? 


In terms of development there is one major issue, namely 
that arrays (accessed by .properties[] ) are saved within the 
SUBFIND folder in HDF5 but aren't necessarily subhalo ordered.
For example, 'Halo_M_Crit200' is a spherical overdensity defined
mass that is calculated around the most bound particle of the 
FOF group (h[0] rather than h[0].sub[0] but currently accessed
only through h[0].sub[0] ) 
The issue is that subhaloes are counted rather than FOF haloes
so the next Halo_M_Crit200 is assigned to h[0].sub[1] (first satellite 
within the first FOF group) rather than h[1] 
The easiest way to identify this is that the properties that should
be passed to FOF haloes ( i.e. h ) have an integer multiple of the
FOF length given by Total_Number_of_groups  (given in FOF attribute
in the HDF5 file) and the properties for the subhaloes are an integer
multiple of Subhalo length given by Total_Number_of_subgroups
(again in the FOF attributes)

Sorry for the horrendous confusion of this... Basically this is a legacy
storage issue. 

Cheers
Alan

apontzen:
No problem. Just to confirm, is this issue you are talking about the same as #258 ? If so, it's probably best dealt with separately as it's a different issue to just tidying up the code which is what I'm doing here.

astroduff:
The fast access is the same as #258 yes, but the issue with the properties that are FOF halo
only being assigned to subhaloes is a bug. However we can close this and open a new one
to deal with that... Especially as it's not an issue that crops up unless you want to use the 
pre-computed property arrays.

Cheers
Alan

apontzen:
Sorry… still unclear: is the bug you are talking about present only in my branch or is it also in the current master branch? If it’s just on this branch we should deal with it here.

Either way could you provide sample code which shows up the error? That would be great.

Thanks!

astroduff:
Apologies for confusing things, the bug is in master too (I'm having trouble actually using older versions of the code that @rokroskar and I first implemented for some reason so perhaps this has always been a stuff up by me). However the basic issue is some properties that are contained within the HDF5 SUBFIND folder (such as Halo_M_Crit200) are assigned to a sub halo when they are instead FOF halos. In the code below the first two FOF halo (given by h[0] and h[1]) have masses very similar to the Halo_M_Crit200 masses that are assigned to the first two subhaloes (given by h[0].sub[0] and h[0].sub[1]) which is reasonable as the read in just grabs anything from SUBFIND folder in the HDF5 file and assumes they are all arrays going to subhaloes. The best way to check this is to compare the size of the array in question and see if it is of size ndim x self.nsubhalos or ndim x self.nfofhalos (where ndim is one in the case of Halo_M_Crit200 for example). 

The exact code issue is:
In [1]: file = '/Users/aduffy/Smaug/Test_NOSN_NOZCOOL_L010N0128/data/subhalos_103/subhalo_103'

In [2]: import pynbody

In [3]: s = pynbody.load(file)

In [4]: h = s.halos()

In [5]: h[0].properties['Mass']
Out[5]: SimArray(28.604694074339932)

In [6]: h[1].properties['Mass']
Out[6]: SimArray(8.880245794949587)

In [7]: h[0].sub[0].properties['Halo_M_Crit200']
Out[7]: SimArray(29.796955896599684)

In [8]: h[0].sub[1].properties['Halo_M_Crit200']
Out[8]: SimArray(8.116568749712314)

In [9]: h[1].sub[0].properties['Halo_M_Crit200']
Out[9]: SimArray(1.4436396431992762)

apontzen:
I see, I think. That sort of sounds like a bug in subfind to me? 

But anyway, for practicalities we might as well put in the workaround. Could you check whether the latest version works for you?

astroduff:
Hey man it's not a bug per se but definitely not ideal... that workaround 
seems to have done the trick though, thanks!

Cheers
Alan

PR #276, opened 2015-08-24
Fix to ramses_util to put corrected star particle birth times in ramses output_xxxx dir

earnric:
This version of the ramses_util.py file creates the birth_xxxx files in the appropriate output_xxxx dir. There was a bug in the previous implementation: the birth files were created in a dir ("birth") as the output directories  -- resulting in (possibly) 100's of thousands of files in a single dir... The code was also not looking for the birth files where ramses_util created them!

This patch ensures the birth files are put in the output directory to which they belong.

Rick


apontzen:
Thanks!

PR #294, opened 2015-11-26
Fix log scaling on colorbar. Add quiverkey background color option.

earnric:
Fixed colorbar scale for log plots -- it was not the log of the value... It is now.
Also added an option to ``def velocity_image`` that allows the user to specify the background color for the quivekey. 

Possible issue with ``n \= 8`` with python 3. Replaced with ``np.divide`` call. Please check this in ramses_util.py to ensure the right type of div is performed (int vs float ... I believe it is supposed to be int)


earnric:
Hum... something is amiss with further testing... Somehow an 'e' is being added to some of the color bar labels. ???
unknown

earnric:
Hum... can figure out why only some of the scalar values in the colorer have an "e" after them. Anyone have more familiarity with Formatters? Seems inconsistent ... (Note I've suppressed the x axis on this plot).

[img_log_Density-z=8.0-0.pdf](https://github.com/pynbody/pynbody/files/45981/img_log_Density-z.8.0-0.pdf)

apontzen:
I have literally never seen this behaviour! Does it still happen with the master version of pynbody, or is it specific to your patch?

earnric:
Only with my patch... Its weird.

earnric:
Ok... I fixed it with a custom formatter ... Things look good now. I think you can pull this... 

unknown

apontzen:
Thanks!

earnric:
You're welcome. :)

PR #317, opened 2016-02-23
Merging in python 3 fix for sound speed

earnric:
Should fix issue with py3 and c_s


earnric:
Opps... wrong direction

PR #334, opened 2016-05-21
A quick fix to ensure the type of MASS is the same as POS (fixes #321)

### LINKED ISSUE #321: bug in gadget.py
Hi, 

I think there is a bug in gadget.py when the mass block is not present and the simulation
was run in single precision.

around line 375 
            # In the header, mass is a double
            block.partlen = 8
            block.data_type = np.float64
            self.blocks[b'MASS'] = block

The mass block should not always be 64 bits, but it should match the type of ,e.g. 'pos'. Otherwise,
there is problems when constructing the kdTree.

Raul

jrs65:
This is a quick fix to ensure that the `mass` SimArray comes out with the same time as `pos`. It seems that if the snapshot has `pos` as single precision, this can break the `kdtree` generation later on (see #321).

This seems to get things working fine though maybe there is a cleaner way to do this.

### DIFF
--- a/pynbody/snapshot/gadget.py
+++ b/pynbody/snapshot/gadget.py
@@ -372,9 +372,10 @@ class GadgetFile(object):
             block = GadgetBlock()
             block.length = 0
             block.start = 0
-            # In the header, mass is a double
-            block.partlen = 8
-            block.data_type = np.float64
+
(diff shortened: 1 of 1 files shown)


apontzen:
Thanks Richard!

Issue #336, opened 2016-05-27
RAMSES runs are missing cells


biernackip:
Hi,

I have been playing with some RAMSES unigrid simulations and found that gas cells are missing. Say for the box with levelmin and levelmax both 7, I should have len(ds.g) to be 2097152 (and I am getting that when loading data with pymses), but pynbody returns instead 1835008.

If I looped through all the possible (x[ii], y[jj]) pairs, where x = np.unique(f.g[‘x’])  and y = np.unique(f.g[‘y’]), and look at the number of cells with these (x, y) coordinates, then this is what I get:

ii even, jj odd : 128
ii even, jj even : 64
ii odd, jj odd : 128
ii odd, jj even : 128

Seems like a bug in the reader to me.

apontzen:
OK, please submit a pull request with the fix! (And provide a sample if possible.)

biernackip:
I wish I knew how to fix it...

perretv:
It seems that the problem concerns all of the ramses outputs.
1/8 of the cells at levelmax are always missing.

apontzen:
I can't reproduce this problem on my outputs. Please post a test file, the unigrid one would be fine.

apontzen:
(Update - no need, seems like @perretv has got a fix)

PR #337, opened 2016-05-30
Ramses levelmax cells issue (issue #336)

perretv:


apontzen:
I think the ngas needs to be updated to 152667 on line 11 of `ramses_test.py`

So to be clear what is going on here - does RAMSES still flag cells at the maximum refinement level for refinement... even though it doesn't actually go on to refine them?

I have never seen this issue in real simulations, but then I have very modest refinement criteria which rarely if ever saturate nlevelmax, so that would make sense.

perretv:
This problem occurs only if the refinements reach levelmax in a given simulation.
Ramses loop over levelmax-1 to build the refinement map, hence this condition in the code.

apontzen:
Thanks!

PR #348, opened 2016-11-16
Projection plots before and after snap.physical_units()

HugoPfister:
For RAMSES, projection plots after doing snap.physical_units() gives different (and wrong) results than the ones obtained with snap.original_units().

This change seems to solve the problem.

### DIFF
--- a/pynbody/plot/sph.py
+++ b/pynbody/plot/sph.py
@@ -408,7 +408,7 @@ def image(sim, qty='rho', width="10 kpc", resolution=500, units=None, log=True,
             im = im / im2
 
     else:
-        im = sph.render_image(sim, qty, width / 2, resolution, out_units=units,
+        im = (sim.properties['boxsize'].in_units(sim['x'].units))**(3-kernel.h_power)* sph.render_image(sim, qty, width / 2, resolution, out_units=units,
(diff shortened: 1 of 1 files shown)


apontzen:
Thanks! Do you understand why this mismatch is arising? It would seem to suggest a deeper problem with the units system that we should fix at source rather than in the sph module. 

An example file/code snippet to reproduce the problem would be helpful to look into this further.

HugoPfister:
Hi,

I ‘tested’ the correction with a sph code and it does not work (since the size of the box is note defined)... I think the units system is okay and the problem comes from a renormalisation in the integration / Kernel2D along the LOS.

Here is a link with a ramses snapshot of a low resolution cosmological simulation. There is also a .py file to generate the 2 snapshots (in the folder) and the mean surface density, which is in both case 0.00873088291913 g cm^-2
https://github.com/HugoPfister/RamsesSim

Let me know if you need something else.

Cheers,
Hugo

On 16 Nov 2016, at 16:08, Andrew Pontzen <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks! Do you understand why this mismatch is arising? It would seem to suggest a deeper problem with the units system that we should fix at source rather than in the sph module.

An example file/code snippet to reproduce the problem would be helpful to look into this further.

PR #353, opened 2017-01-02
Update default_config.ini

smoon24:
As noted in #207, 'vel' should be able to be mapped also with 'Velocities', not only with 'Velocity'.
With this change, GIZMO snapshot file can be properly loaded with 'vel' key.
In addition, the name mapping of 'InternalEnergy' to the key 'u' seems missing.


apontzen:
thanks!

PR #357, opened 2017-02-08
Add CFHTLens SMHMR

bwkeller:
I've added an additional function to pynbody.plot.stars, hudson(), that behaves like behroozi(), but returns the SMHMR from Hudson et al. 2014, one of the only SMHR fits that doesn't rely on abundance matching for halo mass, instead using the CFTLens weak lensing survey to get halo masses.


apontzen:
Thanks!

PR #361, opened 2017-02-21
Ramses snapshots fixes

perretv:
This pull request fixes two minor issues with ramses snapshots:
- A warning message issued by a `f.seek` call
- A recent issue with units due to the use of `omega_b` to assert wether a simulation is cosmological or not

### DIFF
--- a/pynbody/util.py
+++ b/pynbody/util.py
@@ -624,7 +624,7 @@ def read_fortran(f, dtype, n=1):
 def skip_fortran(f, n=1):
     for i in xrange(n):
         alen = np.fromfile(f, _head_type, 1)
-        f.seek(alen, 1)
+        f.seek(alen[0], 1)
         alen2 = np.fromfile(f, _head_type, 1)
         assert alen == alen2
 


pynbody/snapshot/ramses.py
```diff
@@ -794,7 +794,7 @@ def translate_info(sim):
 
     sim.properties['boxsize'] = sim._info['boxlen'] * l_unit
 
-    if sim._info['omega_k'] == sim._info['omega_l'] == sim._info['omega_b'] == 0.0:
```
apontzen: Thanks for the PR - I'm slightly unclear about why checking omega_b is unwanted here?

perretv: Hi,
The default value of `omega_b` in ramses is not zero anymore, so it is not relevant to include it in this test. It actually makes the units cosmological in non-cosmo simulations.

apontzen: OK, weird, but I guess if that's true we have no option... is there a way to detect cosmological simulations done in OCDM (which would have omega_k = omega_l = 0)?

PR #373, opened 2017-05-02
Created read_tipsy_ascii to speed up reading of ascii auxiliary tipsy…

ibackus:
… arrays.

Compared to before, I see about a 4x speedup in ASCII read-time, tested on arrays from 10^5 to 10^8 particles in length.  However, more memory will be used since a full copy of the array needs to be generated.  See below.

This method skips using the slow `numpy.fromtxt` in favor of `readlines`.  Unfortunately, `readlines` reads an approximate amount of text in bytes and therefore we cannot specify the number of lines to read ahead of time.  This means that to perform the mapping from the disk-indexing to the memory-indexing I read the entire array from disk then map its elements to a second array.  This means that during read time double the memory will be required, which is not ideal.  This could be avoid by a smarter interleaving of disk buffers: 1 buffer for `readlines` and 1 buffer for the memory indexing, but that is a little more involved to implement.

### DIFF
--- a/pynbody/snapshot/tipsy.py
+++ b/pynbody/snapshot/tipsy.py
@@ -837,11 +837,7 @@ class TipsySnap(SimSnap):
 
                 # Restart at head of file
                 f.seek(0)
-                f.readline()
-
-            loadblock = lambda count: np.fromfile(
-                f, dtype=dtype, sep="\n", count=count)
-            # data = np.fromfile(f, dtype=tp, sep="\n")
+                
         except ValueError:
             # this is probably a binary file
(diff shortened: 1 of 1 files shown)


apontzen:
Unfortunately I think this swaps CPU complexity for memory complexity and it's not obvious to me that we'll actually help people. Certainly I've been in situations where I'm basically memory-bound with large simulations. If we could have a solution that sticks within the "block" approach that'd be awesome!

ibackus:
Yeah, that's what I was thinking.  It can be done, but it's a little more work.

Issue #375, opened 2017-05-19
Wrong bin range in Fourier analysis of disk profiles


jakobherpich:
In the `pynbody.analysis.profile.Profile` the fourier modes are calculated using binned values for particles azimuth angles. These are calculated with the `np.arctan2` function which maps to the interval [-pi, pi] but the range for the bins used in the code is [0, 2*pi] so half of the disk is not considered for the analysis. I actually do not understand at all, why this gives any reasonable result at all. The lines of code in question are from the file `pynbody/analysis/profile.py`:

```
 620             phi = np.arctan2(
 621                 self.sim['y'][self.binind[i]], self.sim['x'][self.binind[i]])
 622             mass = self.sim['mass'][self.binind[i]]
 623 
 624             x1 = self.sim['x'][self.binind[i]] + self.sim['vx'][self.binind[i]] * delta_t
 625             y1 = self.sim['y'][self.binind[i]] + self.sim['vy'][self.binind[i]] * delta_t
 626             phi1 = np.arctan2(y1,x1) 
 627 
 628             hist, _ = np.histogram(phi, weights=mass, bins=phi_bins, range=(0, 2.*math.pi))
 629             hist1, _ = np.histogram(phi1, weights=mass, bins=phi_bins, range=(0, 2.*math.pi))
 630             binphi = np.linspace(0, 2*math.pi, phi_bins)
 631             for m in range(7):
 632                 f['c'][m, i] = np.sum(hist * np.exp(-1j * m * binphi))
 633                 f['c_delta'][m, i] = np.sum(hist1 * np.exp(-1j * m * binphi))
```

There are two ways to fix this:

1. Adjust the binning range of the histograms to be [-pi, pi].

2. Do the calculation without binning. I chose this option and will create a pull request right now.

apontzen:
@rokroskar could you comment on this and the associated PR? I never used this code so want to double check you agree before I merge the PR :-)

jakobherpich:
The calculation of `binphi` in line 630 is also wrong. These are not the bin centers.

I will implement this option in a new PR as discussed in PR  #376 

> 1. Adjust the binning range of the histograms to be [-pi, pi].

along with a fix to the calculation of `binphi`.

PR #376, opened 2017-05-19
Calculate Fourier modes without binning

jakobherpich:
I changed the method to do the Fourier analysis on non-axisymmetric disk perturbations to avoid loss of information through binning and, thus, increase accuracy. The calculation is based on the method of Quillen et al. 2011 (http://adsabs.harvard.edu/abs/2011MNRAS.417..762Q).

It should also be computationally more efficient because it avoids all the binning and histogram calculations. 

This also fixes issue #375.

I tested the method with multiple galaxies.

### DIFF
--- a/pynbody/analysis/profile.py
+++ b/pynbody/analysis/profile.py
@@ -625,12 +625,9 @@ def fourier(self, delta_t = "0.1 Myr", phi_bins=100):
             y1 = self.sim['y'][self.binind[i]] + self.sim['vy'][self.binind[i]] * delta_t
             phi1 = np.arctan2(y1,x1)
 
-            hist, _ = np.histogram(phi, weights=mass, bins=phi_bins, range=(0, 2.*math.pi))
-            hist1, _ = np.histogram(phi1, weights=mass, bins=phi_bins, range=(0, 2.*math.pi))
-            binphi = np.linspace(0, 2*math.pi, phi_bins)
             for m in range(7):
-                f['c'][m, i] = np.sum(hist * np.exp(-1j * m * binphi))
-                f['c_delta'][m, i] = np.sum(hist1 * np.exp(-1j * m * binphi))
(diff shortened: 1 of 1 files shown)


rokroskar:
actually looks like the binning here was changed a year ago by @apontzen when the 'pattern speed estimation" was added (8fcb42a0). So not sure why the bins are the way they are but I agree that it looks wrong. 

The initial implementation (see c822636) did not use binning. Using the binning speeds up the calculation a lot -- with the binning you only loop through the particles once, whereas without the binning you loop through all the particles for each fourier component. If you have a lot of particles that's a lot of iterations and if you are doing it for just one output you might not care, but if you need to run this for 1000 outputs with 10M particles each then it starts to matter. I found that using ~100 bins or so in phi is sufficient to give nearly identical results to using particles -- I think if I was constructing maps of fourier components I may have used more bins to make it look smoother.

jakobherpich:
I did not realize what @rokroskar said and it makes sense.

I will then change the binning to solve issue #375 and close this PR.

apontzen:
Oh dear, I don't remember doing this. We should add a unit test to make sure it doesn't happen again.

rokroskar:
indeed, a unit test here is sorely lacking

btw, @apontzen did someone request to have the instantaneous pattern speed estimate put in here? Any idea if it matches well with an actual pattern speed measurement?

apontzen:
I think I was just fooling around. It was not used for any science that I know of. Sorry for breaking existing code!

jakobherpich:
@rokroskar No, the pattern speed estimation does not work. One of our students tried to use it and got crazy results.

I'm going to create a different PR that actually implements calculating spectrographs based on multiple snapshots.

apontzen:
we should probably remove the pattern speed estimation in that case, to prevent further confusion... unless the reason it doesn't work is because of the binning error.

jakobherpich:
@apontzen I will test this and if it still fails remove this feature.

PR #377, opened 2017-05-19
Fixes issue #375

jakobherpich:
Replaces PR #376 
The false binning in the calculation of Fourier components is fixed here.

### DIFF
--- a/nose/test_profile.py
+++ b/nose/test_profile.py
@@ -1,6 +1,8 @@
 import pynbody
 import numpy as np
 
+np.random.seed(1)
+
 def make_fake_bar(npart=100000, max=1, min=-1, barlength=.8, barwidth=0.05, phi=0, fraction=0.2):
     
     x = np.random.sample(int(npart*fraction))*(max-min) + min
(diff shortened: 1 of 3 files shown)


rokroskar:
actually changing from distutils to setuptools might not be appropriate here... @apontzen do you have an opinion on this? I just like being able to use `python setup.py develop` which you can't do with distutils...

apontzen:
No major objections but I have always shied away from it, worried that setuptools is not standard. But it seems very rare to find a setup where it is not available any more.

rokroskar:
oops looks like my test fails -- looks like `phi` is actually off by a factor of 2, not sure why at the moment.  I'll try to have a look later.

jakobherpich:
@rokroskar That is expected. `phi` is `m` times the actual angle of the pattern so you'll have to divide the bar speed by m=2.

rokroskar:
doh! I shouldn't do these things late on Friday...

rokroskar:
@apontzen re: distutils vs. setuptools -- should I keep the change in here and merge? I can also undo it if you prefer to stay with distutils.

apontzen:
Thanks both!

PR #378, opened 2017-05-19
Fixes to RAMSES loader, sph plotting with subplot and util

earnric:
**Fixes to sph.py**

- When passing a matplotlib subplot into velocity_image() the code tries to use 'xlim/ylim' to set the plot limits. While this works on 'plot' type objects, if an subplot is passed it it causes an error. Fixed the code to ensure it checks to see if a subplot has been passed in. If so, uses the "set_xlim/ylim" methods on the object.

- Also changed the formatting on the colorbar for plots when LOG is specified. Now the log of the value is taken so the ticks correctly display the log10 of the values displayed.

**Fixes to ramses_util.py**

- the get_tform() routine is used to convert RAMSES format star particle birth dates (via the utility part2birth if necessary). The routine was not setup to work when you load only a subset of CPUs from a RAMSES snapshot. This created a mismatch between the information loaded for gas, particles, etc in the main pynbody.load() mechanism and the get_tform() that crashes the code. I've added an optional parameter that identifies the subset of CPUs you want get_tform to load.

**Fixes to util.py**

- when reading in FORTRAN binary data, the f.seek() call was using an array as a scalar -- causing a warning. I fixed this. Since the array is always 1 element long, it now calls f.seek(alen[0], 1)...


pynbody/analysis/ramses_util.py
```diff
@@ -398,7 +398,7 @@ def write_ahf_input(sim, tipsyfile):
     f.close()
 
 
-def get_tform(sim, part2birth_path=part2birth_path):
+def get_tform(sim, part2birth_path=part2birth_path, cpuRange=None):
```
earnric: Handle cpu ranges in ramses load

pynbody/plot/sph.py
```diff
@@ -464,13 +471,29 @@ def image(sim, qty='rho', width="10 kpc", resolution=500, units=None, log=True,
...
 
-        if units.latex() is "":
-            units=""
-        else:
-            units = "$"+units.latex()+"$"
-
+        # RS - Fix formatting.
         if show_cbar:
-             plt.colorbar(ims).set_label(qtytitle+"/"+units)
+            if log:
```
earnric: Use log10(value) in the colorbar if the colorbar is on a log scale... saves space and agrees with labels like "log \rho"

pynbody/analysis/ramses_util.py
apontzen: Could we make this cpu_range instead of cpuRange? PEP 8 compliance

pynbody/plot/sph.py
```diff
@@ -214,7 +221,7 @@ def volume(sim, qty='rho', width=None, resolution=200,
         ctf.add_rgb_point(vmin+(vmax-vmin)*0.8,200./255,178./255,164./255)
         ctf.add_rgb_point(vmin+(vmax-vmin)*0.9,1.0,210./255,149./255)
         ctf.add_rgb_point(vmax,1.0,222./255,141./255)
-        print vmin,vmax
+        print (vmin,vmax)
```
apontzen: Actually I'm not sure this print statement should be here at all?

pynbody/plot/sph.py
```diff
@@ -464,13 +471,29 @@ def image(sim, qty='rho', width="10 kpc", resolution=500, units=None, log=True,
...
+
+            if qtytitle is not None:
+                plt.colorbar(ims,format=l_f).set_label(qtytitle)
+            else:
+                if log:
+                    plt.colorbar(ims,format=l_f).set_label("$log\; %s$" % units)
+#                    plt.colorbar(ims,
+                else:
+                    plt.colorbar(ims,format=l_f).set_label(units)
+
```
apontzen: I am worried about these changes. Please see comments on revision 5acf3411b3d101b4df, which this partially (or fully) reverts. I kept getting really bizarre results on my plots and I think this complicated logic may only work for a small number of cases.

pynbody/analysis/ramses_util.py
earnric: Done...

pynbody/plot/sph.py
earnric: Removed the print

pynbody/plot/sph.py
earnric: ok, understand and I've reverted this so there's not change.

apontzen:
Thanks!

Issue #385, opened 2017-06-21
units bug in half_light_r


ferahm:
There is a units bug in pynbody.analysis.luminosity.half_llight_r the output thinks it's in units of distance squared.  The value is correct, but the units are not.

jakobherpich:
The reason is in the last line of the function, when `test_r` which is a SimArray with the correct unit is multiplied by the unit again. I suspect this broke, when we changed the behavior of `np.max()` and `np.min()` to return the units of its argument when the argument is a SimArray. I'll create a pull request to fix this.

PR #386, opened 2017-06-21
Fixes issue #385

jakobherpich:
The squared units from #385 are corrected.

If it really is due to the changed behavior of `np.max()` and `np.min()` in pynbody this bug has been around for a long time. How come no one has noticed it yet?

### DIFF
--- a/pynbody/analysis/luminosity.py
+++ b/pynbody/analysis/luminosity.py
@@ -155,4 +155,4 @@ def half_light_r(sim, band='v'):
         else:
             min_low_r = test_r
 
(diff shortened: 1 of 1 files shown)


rokroskar:
make sure you add a unit test and try to make sure there are no other untested spots that this change will now break

jakobherpich:
Unfortunately, I won't have time to do this soon. I'll be offline for the next few weeks and am currently dealing with other issues, that occured when I tried to fix this. Can someone else write a test? @rokroskar?

PR #401, opened 2017-07-22
render stars new features

TobiBu:
I added a new feature to the pynbody.plot.stars.render function to include simple dust absorption estimation via Calzetti's law.

I further added a new function pynbody.plot.stars.render_mollweide which uses the pynbody.sph.render_spherical function plus healpy's mollview function to render an RGB all-sky view of the stars in three different bands.


apontzen:
Looks nice to have, but it's broken the setup. I don't think we should be modifying the dependencies, since many people would not be using this feature and still want the basic pynbody functionality.

So I'd suggest reverting setup.py to the original form and making sure the code fails gracefully if the required dependencies are not installed.

Thanks!

TobiBu:
I have removed the dependencies from setup.py and do the dependency checking in the routine itself and in case of uninstalled packages print a warning.

apontzen:
Thanks!

Issue #379, opened 2017-05-26
reading 3D tipsy auxiliary arrays


ibackus:
I'm using ChaNGa which can save arrays for particle accelerations which are 3D quantities (i.e. require arrays shape Nx3) and I save them in the tipsy format.  Pynbody does not seem to be able to read these as 3D arrays, instead it just reads the first N elements.  Is there a way to specify the quantity is 3D?

#### example
Generate auxiliary arrays required
```ChaNGa -n 0 snapshot.param```
Read in pynbody
```python
f =  pynbody.load('snapshot.000000')
print f['acc2'].shape
 >>> (40412,)
```
Which should be `(40412, 3)`

apontzen:
Sadly no I don't think this is currently supported. Shouldn't be too complicated to add though.

PR #403, opened 2017-08-11
Allow reading of arbitrary dimension tipsy aux arrays (e.g. 1D or 3D)

ibackus:
This addresses issue #379.

This is unfortunately not straightforward, since the tipsy format does not include information about the number of columns in a tipsy array.

There are several different file types, all which have dimensions inferred
using different methods:

 * For binary arrays, the number of dimensions are inferred robustly from the
file size, dtype, and number of particles.
 * For gzipped binary arrays, the uncompressed filesize cannot be directly
read for files >4GB, but in general it is possible to correctly estimate the
number of dimensions
 * For ascii arrays, the number of dimensions are estimated by estimating the
average size (in bytes) of a single data point and dividing the file size
by this and the number of particles.  This should be fairly robust except in
some edge cases.  A warning is raised if the estimate appears bad.
 * Reading of gzipped ASCII arrays has also been implemented, but the number
of dimensions cannot be estimated

### DIFF
--- a/pynbody/snapshot/tipsy.py
+++ b/pynbody/snapshot/tipsy.py
@@ -793,9 +793,6 @@ class TipsySnap(SimSnap):
             except IOError:
                 pass
 
-        import sys
-        import os
-
         # N.B. this code is a bit inefficient for loading
         # family-specific arrays, because it reads in the whole array
         # then slices it.  But of course the memory is only wasted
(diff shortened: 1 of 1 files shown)


ibackus:
I'm not sure why this has failed a test.  It has failed the python 3.6 test, with the error message:
```python
File "/home/travis/miniconda/envs/3.6_1.13.1/lib/python3.6/site-packages/pynbody-0.41-py3.6-linux-x86_64.egg/pynbody/snapshot/tipsy.py", line 899, in __read_array_from_disk
    array_shape = [n_read, ndim]
UnboundLocalError: local variable 'ndim' referenced before assignment
```
However, it should not be able to get to that point in the code without `ndim` being assigned a value.  This is especially bizarre, because all the other test builds are passed.

PR #407, opened 2017-09-20
fix the bug on vr calculation in derived.py

hellkonig:
In original derived.py, the way to calculate radial velocity vr is

vr = (vx*x + vy*y + vz*z) / r

However, it is not true, the correct way should be

vr = sqrt(vx^2*x^2 + vy^2*y^2 + vz^2*z^2) / r.

I replace the old code 
`return (self['pos'] * self['vel']).sum(axis=1) / self['r']`
by
`return np.sqrt(((self['pos'] * self['vel'])**2).sum(axis=1)) / self['r']`
in line 41 of derived.py

### DIFF
--- a/pynbody/derived.py
+++ b/pynbody/derived.py
@@ -38,7 +38,7 @@ def rxy(self):
 @SimSnap.derived_quantity
 def vr(self):
     """Radial velocity"""
-    return (self['pos'] * self['vel']).sum(axis=1) / self['r']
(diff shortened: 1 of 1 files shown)


apontzen:
Could you cite your source/give a reason for this? For most purposes the natural meaning of radial velocity I'd expect to be dr/dt.

As d r^2/dt = 2(vx x + vy y + vz z), one gets the result in the original code.

hellkonig:
Oh, yes, you are right. Sorry for the mistake I made.

apontzen:
No problem, thanks for confirming

PR #408, opened 2017-10-02
Hop support

Martin-Rey:
Interfacing HOP group catalogue to be used transparently through pynbody.


apontzen:
Uh oh! Looks like the "old numpy" version has broken. But I think this relates to other package intercompatibilities. It's fair for us to stop supporting numpy 1.7 now, so maybe just strip that line from the .travis.yml?

PR #411, opened 2017-10-11
Merger tree

TobiBu:
This is a commit to handle merger tree files and plot merger trees with pynbody.
I created a file "tree.py" containing classes to handle files created with the AHF merger tree tool and to read in merger trees and halo properties from the AHF halo catalogue.
There is also an abstract class to implement more merger tree tool like e.g. the rockstar consistent-tree tool.

This commit is still laking a written documentation.

### DIFF
--- a/pynbody/configuration.py
+++ b/pynbody/configuration.py
@@ -56,6 +56,10 @@ def _get_basic_config_from_parser(config_parser):
                                         config_parser.get('general', 'snap-class-priority').split(","))
     config['halo-class-priority'] = map(str.strip,
                                         config_parser.get('general', 'halo-class-priority').split(","))
+    config['merger-tree-priority'] = map(str.strip,
+                                        config_parser.get('general', 'merger-tree-priority').split(","))
+    config['merger-tree-history-priority'] = map(str.strip,
+                                        config_parser.get('general', 'merger-tree-history-priority').split(","))
 
 
     config['default-cosmology'] = {}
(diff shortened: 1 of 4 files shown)


apontzen:
Could you add an example so I can test this in real-world use? Thanks!

TobiBu:
You mean just a piece of code showing how it works?
I also thought about writing a short tutorial if it gets merged.

apontzen:
Yes a piece of code or even better a tutorial. It's hard at the moment to verify what it does. Thanks!

TobiBu:
Hi,
I created a jupiter notebook explaining the basic functions. If you have AHF installed and the merger tree tool configure (in the subfolder analysis in the AHF folder) you should be able to run the notebook on one of your galaxies. Please let me know if everything works for you. On my laptop it works with python3 and on the server with python2.7. So there should be no problems. 

[merger_tree_tutorial.ipynb.zip](https://github.com/pynbody/pynbody/files/1556469/merger_tree_tutorial.ipynb.zip)

apontzen:
Sorry for long delay - I think given where tangos has got to, this probably doesn't belong in pynbody. Do you agree or should some aspect of this functionality still be merged?

TobiBu:
No, I think the tangos implementation should be sufficient.
The only thing which still might be interesting is the automatic run of the AHF MergerTree code in case the mtree files are not present...

apontzen:
Sounds like it could be a helpful script but again probably belonging to tangos rather than pynbody

PR #421, opened 2017-11-13
Grafic python3 compatibility and loading of ic_refmap/pvar files

Martin-Rey:
I add some compatibility issues when using Grafic snaps with python3. As far as I can test, these changes should fix it.

In the process, I added loading of extra IC files produced for RAMSES: ic_refmap and ic_pvar0000*.

Let me know if this breaks anything.

### DIFF
--- a/pynbody/snapshot/grafic.py
+++ b/pynbody/snapshot/grafic.py
@@ -5,7 +5,7 @@ grafic
 
 Support for loading grafIC files
 """
-
+from __future__ import print_function
 from .. import util
 from .. import array
 from .. import chunk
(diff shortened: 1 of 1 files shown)


apontzen:
If using the print function, should add `from __future__ import print_function` at the top of the file to maintain python 2.x compatibility. Otherwise looks good!

Ideally we'd add some units tests for grafic but this can wait

Martin-Rey:
Just added it. 

Out of curiosity, why would the code run flawlessly as far as I could test without the import. Why is it not giving a Syntax Error at compile time ?

apontzen:
e.g. Python 2 without the future import will interpret `print(a,b)` as the statement `print` followed by the tuple `(a,b)` so syntactically it's fine but the behaviour is different.

Issue #11, opened 2013-06-01
SimArray handling with numpy functions


apontzen:
_Original author: rokroskar (December 13, 2010 10:56:52)_

this doesn't work:

&gt;&gt;&gt; np.sum(s['mass'])

this does:

&gt;&gt;&gt; s['mass'].sum()

this gives the wrong units:

&gt;&gt;&gt; np.sqrt((s['pos'][:,0:3]**2).sum(axis=1))
SimArray([ 150.93175958,   67.12237074,    7.0066101 , ...,    6.47421208,
         10.23414716,   10.68530362], 'kpc**2')

this gives the correct units:

&gt;&gt;&gt; (s['pos']**2).sum(axis=1)**(1,2)
SimArray([ 150.93175958,   67.12237074,    7.0066101 , ...,    6.47421208,
         10.23414716,   10.68530362], 'kpc')

It would be nice if these sets of operations gave the same results, otherwise this is bound to lead to mayhem...

_Original issue: http://code.google.com/p/pynbody/issues/detail?id=11_

Issue #10, opened 2013-06-01
Make the sample code on the front page work


apontzen:
_Original author: andrew.p...@gmail.com (December 10, 2010 11:08:52)_

We should implement pynbody.plot.rotation_curve so that our front-page example generates a gorgeous plot every time ;)

Rok?

_Original issue: http://code.google.com/p/pynbody/issues/detail?id=10_

PR #423, opened 2017-11-17
Update snapshot.ramses to work with new format

apontzen:
- explicitly named and typed arrays
 - particle families are now explicit

TODO: add test for new format - should continue to work with old format


biernackip:
There seem to be a problem - some mismatch in length by factor of 8. I am getting this:
```
Python 2.7.6 | 64-bit | (default, Sep 15 2014, 17:36:10)
Type "copyright", "credits" or "license" for more information.

IPython 5.3.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import pynbody

In [2]: ds = pynbody.load('output_00001/')
/store/uzh/uzh5/Enthought/Canopy_64bit/User/lib/python2.7/site-packages/pynbody/snapshot/ramses.py:374: RuntimeWarning: RamsesSnap is configured to use multiple processes, but the posix_ipc module is missing. Reverting to single thread.
  warnings.warn("RamsesSnap is configured to use multiple processes, but the posix_ipc module is missing. Reverting to single thread.", RuntimeWarning)
/store/uzh/uzh5/Enthought/Canopy_64bit/User/lib/python2.7/site-packages/pynbody/snapshot/ramses.py:521: UserWarning: Using field 'family' to distinguish stars. If this is wrong, try editing your config.ini, section [ramses], entry particle-distinguisher.
  warnings.warn("Using field %s to distinguish stars. If this is wrong, try editing your config.ini, section [ramses], entry particle-distinguisher."%pb_name)
IOError                                   Traceback (most recent call last)
<ipython-input-2-e51766571f9b> in <module>()
----> 1 ds = pynbody.load('output_00001/')

/store/uzh/uzh5/Enthought/Canopy_64bit/User/lib/python2.7/site-packages/pynbody/snapshot/__init__.pyc in load(filename, *args, **kwargs)
   1945         if c._can_load(filename):
   1946             logger.info("Loading using backend %s" % str(c))
-> 1947             return c(filename, *args, **kwargs)
   1948
   1949     raise IOError(

/store/uzh/uzh5/Enthought/Canopy_64bit/User/lib/python2.7/site-packages/pynbody/snapshot/ramses.pyc in __init__(self, dirname, **kwargs)
    405         self._maxlevel = kwargs.get('maxlevel', None)
    406
--> 407         ndm, nstar = self._count_particles()
    408
    409         has_gas = os.path.exists(

/store/uzh/uzh5/Enthought/Canopy_64bit/User/lib/python2.7/site-packages/pynbody/snapshot/ramses.pyc in _count_particles(self)
    528                              [self._particle_filename(i) for i in self._cpus],
    529                              [distinguisher_field]*len(self._cpus),
--> 530                              [distinguisher_type]*len(self._cpus))
    531
    532         for npart_this, nstar_this, my_mask in results:

/store/uzh/uzh5/Enthought/Canopy_64bit/User/lib/python2.7/site-packages/pynbody/snapshot/ramses.pyc in remote_map(*args, **kwargs)
     68
     69     def remote_map(*args, **kwargs):
---> 70         return map(*args[1:], **kwargs)
     71
     72

/store/uzh/uzh5/Enthought/Canopy_64bit/User/lib/python2.7/site-packages/pynbody/snapshot/ramses.pyc in q(*args)
     63         def q(*args):
     64             t0 = time.time()
---> 65             r = fn(*args)
     66             return r
     67         return q

/store/uzh/uzh5/Enthought/Canopy_64bit/User/lib/python2.7/site-packages/pynbody/snapshot/ramses.pyc in _cpui_count_particles(filename, distinguisher_field, distinguisher_type)
     94     try:
     95         skip_fortran(f, distinguisher_field)
---> 96         data = read_fortran(f, distinguisher_type, header['npart'])
     97     except TypeError:
     98         data = np.zeros(npart_this)

/store/uzh/uzh5/Enthought/Canopy_64bit/User/lib/python2.7/site-packages/pynbody/util.pyc in read_fortran(f, dtype, n)
    610     if alen != length:
    611         raise IOError, "Unexpected FORTRAN block length %d!=%d" % (
--> 612             alen, length)
    613
    614     data = np.fromfile(f, dtype, n)

IOError: Unexpected FORTRAN block length 645!=5160
```

biernackip:
Also - cloud particles are not bh/sink particles. These are only helper particles internally for ramses. Sink particle is stored separately - I can provide more info about the format, but I think it should be a separate branch.

apontzen:
That looks like you have not installed the new branch, as your trace back has old code in it

biernackip:
Oops! Sorry! I have reinstalled and looks like it works fine.

The only thing: I would rename `bh` to `clouds`. Do you think it would be hard to read proper sinks (which are used as both black holes and stars) from a CSV file? It's also in the ramses output dir (`output_0????/sink_0????.csv`)

apontzen:
OK, good. No problem to rename the particle family to be more accurate.

Why are sinks in a separate CSV file? Is there any way this can be changed? If not I can look at reading them in but it implies even more conditional pathways and so the reader code gets more and more intricate and hard to follow...

biernackip:
I see at least few reasons:
1. Their behaviour is different than all other particles
2. It might be less handy to analyse them if they are part of large bundle (if one is interested in a quick analysis)
3. Each CPU has a copy of all the sinks to speed up certain calculations
4. They can be used as either BH or stars
5. Bad design choice?

biernackip:
@apontzen for sink particles: would you prefer a format similar to particles (dumped as a separate binary file with a separate file descriptor) or a simple csv file with first row describing quantities?

apontzen:
On balance, probably the CSV file. If it can't be integrated with the rest of the particle data, there needs to be special handling anyway and then the CSV file will be easier to read (and presumably there will only be a few particles even for a large volume, so speed will not be a big problem?)

biernackip:
OK, great. This PR works fine, as far as I can tell. Shall we open another issue for sinks then?

apontzen:
Ideally I'd like to add a file to the test data archive so we can explicitly test the new format in Travis builds. So it'd probably be worth addressing the sinks in the same issue and updating it all at once.

biernackip:
Sounds good. Can the format be something like:
#idsink,msink,xsink,ysink,etc...
#-,mass,pos,pos,...
1,0.1,0.5,0.5,...

Where the first line describes quantity, the second one the kind of quantity and the following ones are showing values in RAMSES code units.

Should we have a predefined set of quantities or is it possible to load them all as long as the unit/kind makes sense?

apontzen:
I'm not sure I totally get the distinction between the first and second lines. Preferably call the arrays the same as they are called in the standard particle blocks (where there is an equivalent), so that only one name mapping setup is required.

That would be something more like:

```
# identity, mass, position_x, position_y, position_z, ...
```

If you are feeling generous you could also add a line to hint at a suitable datatype, e.g.

```
# int, float, float, float, float, ...
```

biernackip:
Everything except of the ID would be a float. I was thinking about storing more quantities (like sound speed nearby, density of gas in surrounding, etc.), the way they are calculated internally in RAMSES will for sure be different than estimate from pynbody. This also would be in line with what people do in e.g. AREPO, where they also store all these quantities.

luturonunca:
hello I am getting an issue when pynbody is trying to load the SIM, specifically when is trying to read the part_file_descriptor.txt file. In my version of Ramses (maybe Im not running the right commit ) this looks like this:
` version:            1
variable # 1: position_x
variable # 2: position_y
variable # 3: position_z
variable # 4: velocity_x
variable # 5: velocity_y
variable # 6: velocity_z
variable # 7: mass
variable # 8: identity
variable # 9: levelp
variable #10: family
variable #11: tag
`

So when loading with pynbody the following error comes due to an inconsistency in the formats:

`Python 2.7.12 (default, Nov 19 2016, 06:48:10) 
Type "copyright", "credits" or "license" for more information.

IPython 5.1.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import pynbody

In [2]: ds = pynbody.load('output_00041')
/usr/local/lib/python2.7/dist-packages/pynbody-0.41-py2.7-linux-x86_64.egg/pynbody/snapshot/ramses.py:425: RuntimeWarning: RamsesSnap is configured to use multiple processes, but the posix_ipc module is missing. Reverting to single thread.
  RuntimeWarning)
-------------------------------------------------------------------
ValueError                        Traceback (most recent call last)
<ipython-input-2-7303840957bf> in <module>()
----> 1 ds = pynbody.load('output_00041')

/usr/local/lib/python2.7/dist-packages/pynbody-0.41-py2.7-linux-x86_64.egg/pynbody/snapshot/__init__.pyc in load(filename, *args, **kwargs)
   1945         if c._can_load(filename):
   1946             logger.info("Loading using backend %s" % str(c))
-> 1947             return c(filename, *args, **kwargs)
   1948 
   1949     raise IOError(

/usr/local/lib/python2.7/dist-packages/pynbody-0.41-py2.7-linux-x86_64.egg/pynbody/snapshot/ramses.pyc in __init__(self, dirname, **kwargs)
    378         self._filename = dirname
    379         self._load_infofile()
--> 380         self._setup_particle_descriptor()
    381 
    382         assert self._info['ndim'] <= 3

/usr/local/lib/python2.7/dist-packages/pynbody-0.41-py2.7-linux-x86_64.egg/pynbody/snapshot/ramses.pyc in _setup_particle_descriptor(self)
    476     def _setup_particle_descriptor(self):
    477         try:
--> 478             self._load_particle_descriptor()
    479         except IOError:
    480             self._guess_particle_descriptor()

/usr/local/lib/python2.7/dist-packages/pynbody-0.41-py2.7-linux-x86_64.egg/pynbody/snapshot/ramses.pyc in _load_particle_descriptor(self)
    489             for l in f:
    490                 if not l.startswith("#"):
--> 491                     ivar, name, dtype = map(str.strip,l.split(","))
    492                     self._particle_blocks.append(self._translate_array_name(name, reverse=True))
    493                     self._particle_types.append(dtype_mappings[dtype])

ValueError: need more than 1 value to unpack`

So is it the version i am running of Ramses or is it pynbody?

thanks

apontzen:
I guess it's the version of ramses which has an unfinished specification for the part_file_descriptor - @biernackip could you comment?

luturonunca:
yes you were right, latest commit of Ramses has another format thanks.

luturonunca:
helle again.

I have another issue. It seems that the particles Ids are not being retrieved correctly. I am getting a allzeros array for the Ids.
like this:

```
Python 2.7.13 (default, Jan 28 2017, 17:29:29) 
Type "copyright", "credits" or "license" for more information.

IPython 5.0.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import pynbody

In [2]: ini = pynbody.load('output_00001')
/usr/lib64/python2.7/site-packages/pynbody-0.41-py2.7-linux-x86_64.egg/pynbody/snapshot/ramses.py:425: RuntimeWarning: RamsesSnap is configured to use multiple processes, but the posix_ipc module is missing. Reverting to single thread.
  RuntimeWarning)

In [3]: final = pynbody.load('output_00041')

In [4]: ini['iord']
Out[4]: SimArray([0, 0, 0, ..., 0, 0, 0])

In [5]: final['iord']
Out[5]: SimArray([0, 0, 0, ..., 0, 0, 0])

```

apontzen:
Thanks @luturonunca - this seems to be an issue where pynbody was expecting int64 but actually the integer fields are stored as int32. I've fixed it just by changing pynbody's interpretation of "i" in the particle descriptor file. Looking at the bitbucket repository for dump_utils.f90 it seems that "q" would actually indicate int64 - is that correct?

Taking this opportunity to find out a bit more - why is it that the IDs are emerging as int32?  My understanding is that internally the ramses IDs are in fact 64-bit (i8b?) so why do they get down-cast to int32? Or have I misunderstood? (Context: for a project with @martin-rey it would be very helpful to use 64-bit IDs.)

cphyc:
@apontzen the reason IDs are dumped as 32bit integers is because of the line in `amr_parameters.f90` that reads:
`#ifndef LONGINT
  !integer,parameter::i8b=4  ! default long int are 4-byte int
  integer,parameter::i8b=selected_int_kind(9) ! since log(2*10^9)/log(2)=30.9
#else
  integer, parameter :: i8b=selected_int_kind(18) ! since log(2*10^18)/log(2)=60.8
  !integer,parameter::i8b=8  ! long int are 8-byte int
#endif /* LONGINT */
`

apontzen:
@cphyc Ah thanks! Paging @Martin-Rey

cphyc:
@apontzen @Martin-Rey BTW, if you use more than `2**31` points, you'll need longer integer, which you can achieve using the compilation flag `-DLONGINT` (you'll get 64bit integers then).

Martin-Rey:
Thanks @cphyc.  The output type is indeed int64 when compiling with LONGINT.

However, I am still getting the same as @luturonunca, e.g. zeroes everywhere when loading any array, both for integers and float/double.

screen shot 2017-11-30 at 16 38 24

apontzen:
@martin-Rey weird! Do the unit tests pass?

Martin-Rey:
Nope, I get failures for particle loading in both old and new format:
FAIL: ramses_test.test_particle_arrays
FAIL: ramses_new_ptcl_format_test.test_load_iord
FAIL: ramses_new_ptcl_format_test.test_load_pos
because it loads zeroes rather than the values.

The other issue for me is when using old format snapshots. For some reason the threading for reading/image processing does not operate anymore and everything is done on one thread.

Everything is nominal when checking out of this branch.

apontzen:
Very strange. Unfortunately I can't reproduce this issue and as you can see the unit tests pass on the Travis setups. There must be a specific library version or other aspect of your setup that this branch does not like; are you able to isolate it?

apontzen:
Hi @biernackip, just checking in to see if there is an updated test snapshot available with BH particles? Would be good to get this closed before the holidays.

apontzen:
Note to self: we've discovered this patch no longer works with posix_ipc multi-processing reading.

To do:
 * [ ] Fix
 * [ ] Add unit test or Travis configuration to catch this in future

biernackip:
Hi @apontzen et all,
Sorry for late reply, I was travelling almost the whole December. I am closing on the sink format (one csv file per snapshot). It's in testing now, once it passes, I will send an output for testing (hopefully Monday).

biernackip:
Here is the proposed descriptor for sinks:
```
# id,msink,x,y,z,vx,vy,vz,lx,ly,lz,tform,acc_rate,rho_gas,cs**2,etherm,vx_gas,vy_gas,vz_gas,mbh
# 1,m,l,l,l,l t**-1,l t**-1,l t**-1,m l**2 t**-1,m l**2 t**-1,m l**2 t**-1,t,m t**-1,m l**-3,l**2 t**-2,m l**2 t**-2,l t**-1,l t**-1,l t**-1,m
```
where:
l - length
t - time
m - mass
and 1 is for unitless

apontzen:
@biernackip could you test the latest version of this branch? Should now be working with sinks (by default these make "bh" particles).

apontzen:
@Martin-Rey would be good if you could check this now works for you - the problem with posix-ipc should be fixed.

biernackip:
@apontzen this is amazing! Works like a charm! Is the sink information used in potential calculations?

The only thing I get with these outputs are some warning (h, aexp, OmegaM and OmegaB are defined in the info file...)

```
/home/cluster/pbiern/python/lib/python2.7/site-packages/pynbody-0.41-py2.7-linux-x86_64.egg/pynbody/simdict.py:92: RuntimeWarning: Assuming default value for property 'a'=1.00e+00
  name, value), RuntimeWarning)
/home/cluster/pbiern/python/lib/python2.7/site-packages/pynbody-0.41-py2.7-linux-x86_64.egg/pynbody/simdict.py:92: RuntimeWarning: Assuming default value for property 'h'=6.78e-01
  name, value), RuntimeWarning)
/home/cluster/pbiern/python/lib/python2.7/site-packages/pynbody-0.41-py2.7-linux-x86_64.egg/pynbody/simdict.py:92: RuntimeWarning: Assuming default value for property 'omegaM0'=3.09e-01
  name, value), RuntimeWarning)
/home/cluster/pbiern/python/lib/python2.7/site-packages/pynbody-0.41-py2.7-linux-x86_64.egg/pynbody/simdict.py:92: RuntimeWarning: Assuming default value for property 'omegaL0'=6.91e-01
  name, value), RuntimeWarning)
```

apontzen:
@biernackip great! I just pushed (hopefully) a fix to this residual problem - let me know if this solves it for you.

biernackip:
OK, warnings are gone, but now whatever simulations I try (that is not comological) and whatever output I try, I get the same time
```
In [12]: ds.properties['time'].in_units('Myr')
Out[12]: 651875.9674711871
```

apontzen:
I think there is some confusion about how to tell a cosmological from a non-cosmological ramses output. At the moment it is checking whether omega_l=omega_k=omega_b=0, and if yes it assumes non-cosmological. So I guess your outputs have omega_b!=0? 

What is the definitive way to tell?

Martin-Rey:
@apontzen Looks like the new format branch is working on my setup, both with and without posix_ipc.

biernackip:
@apontzen I think I would just limit the check to omega_l=omega_k=0

There must have been a change a bit ago (there was a PR from Valentin Perret, if I recall correct) and now the omega_b is stored. We can brainstorm some better way of doing it... 

Since a long while (~3 yrs) there is the `namelist.txt` file stored in the each output directory. In `RUN_PARAMS` there is a `cosmo=.false./.true.` flag and if not, then by default it is also `.false.`. This would of course require reading one more file...

apontzen:
OK. I looked at some namelist.txt examples from recent runs and sometimes they appear garbled (?!) so for now I have just implemented omega_l=omega_k=0. Let me know if this solves it for you.

biernackip:
Unfortunately, this didn't solve it :/ (I cleaned the `build` dir and python site-packages entry)

They are garbled when RAMSES is compiled without `-assume byterecl` (for intel compiler). This part of RAMSES is, sadly, very compiler dependent

apontzen:
Can you figure out why this hasn't solved it? I am slightly in the dark...

biernackip:
It seems to me like the condition is switched. If `self._info['omega_k'] == self._info['omega_l'] == 0` then it's non-cosmological, now the condition at line 997 implies the opposite (I'd just write in 997: `if sim._looks_cosmological():` (and maybe rename `_looks_cosmological` to `_not_cosmological`)

biernackip:
@apontzen Do you think this can be now merge to the master branch?

apontzen:
I think so - there was some weird issue where the Travis build of the most recent merge hung but I can't reproduce it now...

biernackip:
How does it handle the case if the sink file is empty (no sinks formed)? An error? A warning? Nothing? Empty array?

apontzen:
I suggest it's worth checking this (could you report back?). The correct behaviour should be that `f.bh` would be empty (`len(f.bh)==0`). If possible I guess we should add a unit test -- could do it on the 'old format' file. Not an ideal test, but better than nothing.

biernackip:
Yup, returns zero length. 

However, it would be good to have a check if the sink is not an old generation file. So if the first two lines are not comments, then pynbody should skip reading the sink file. In such a case it would be the best to have a warning that old sinks are not supported and have an empty array.

biernackip:
Tested. Works perfectly!

biernackip:
Getting circular dependency when I try to check the `tform` of stars with this commit.

PR #424, opened 2017-11-17
Avoiding setting gas mass to zero when loading AMR cells for DM only sims

Martin-Rey:
When loading AMR cells with force_gas=true for DM only runs, snap.gas['rho'] and hence snap.gas['mass'] are set to zero. This yields divisions by zero when plotting gas quantities for example.

This is a quick fix to ensure that these arrays are non zero and such that snap.gas['mass']=1.

### DIFF
--- a/pynbody/snapshot/ramses.py
+++ b/pynbody/snapshot/ramses.py
@@ -599,6 +599,14 @@ class RamsesSnap(SimSnap):
             dims.append(self.gas[i])
             self.gas[i].set_default_units()
 
+
+        if not os.path.exists(self._hydro_filename(1)) and self.gas['rho'].all() == 0:
+            #Case where force_gas = True, make sure rho is non-zero and such that mass=1
+            logger.info("No hydro file found, gas likely from force_gas=True => hard setting rho gas")
+            self.gas['rho'] += 1.0
+            self.gas['rho'] /= (np.array(self.gas['smooth']) ** 3)
+
+
         nvar = len(dims)
 
         grid_info_iter = self._level_iterator()


apontzen:
Thanks @Martin-Rey - just pausing this merge until #423 is merged first

biernackip:
This feels unsafe. I think snap.gas should just not exist and then throw warnings/errors when you try to access it with plotting.

apontzen:
It's not new - this has always been available by setting the same flag (`force_gas`), so that the AMR structure can be investigated. Unfortunately without reasonable mass and rho arrays the interpolated images then fail, so I support htis patch.

Martin-Rey:
I don't understand why the last test does not pass in Travis. The added ramses test works on my local machine both with and without posix_ipc. 

Can you reproduce this behaviour ?

apontzen:
It's always frustrating when Travis builds can't easily be recreated - don't have particular advice here other than persistence. Have you tried on both linux and mac os?

apontzen:
By the way, travis does offer a debug mode where you can ssh the build -- it's a bit of a last resort though as it's often deeply frustrating. https://docs.travis-ci.com/user/running-build-in-debug-mode/

Martin-Rey:
I get extremely weird behaviour from this PR now that I have started to look into it. 

Attached is an extract from an ipython session in which reloading the simulation multiple times yields different results for the arrays mass and rho (derived, not loaded from disks in this case).

Have you ever seen this behaviour ?
screen shot 2018-02-15 at 15 51 46

I cannot reproduce this issue outside of ipython and cannot find the configuration that makes the Travis build fail, neither on linux or Mac. It is really hard to understand as the Travis test fail without posix_ipc but work with it...

Martin-Rey:
Alright, changing the if condition seems to have made the behaviour more reliable.

The tests fail because of version compatibility now, between Sphinx1.7 and old ipython I think ( see https://github.com/ipython/ipython/issues/10666).

This is probably more general than just this PR.

Martin-Rey:
Sorry for the commit spam. In short, sphinx v1.7 has deprecated a few packages from v1.6 which was causing a problem in the documentation build.

However, sphinx v1.6 is not compatible with old python 3 versions (namely python 3.4). The workaround is to install sphinx v1.6 when building the documentation as this is done only with python2 versions.

Let me know if you don't like this solution.

PR #435, opened 2018-02-02
Fix to ramses birth times

biernackip:
Format of files produced by part2birth in RAMSES has changed (don't really know when). I tried to use current method (`np.fromfile`), but couldn't ever get a right count. Works like a charm with `fortranfile.py` and its classes (similar routine is also part of scipy, but as this is not a library required by pynbody, I decided to include additional file - perhaps there is an easier way to do it).

This has been tried on a RAMSES cosmo simulation that requires new file format for particles, but should not depend on that, as long as `part2birth` output is reasonable.


apontzen:
Thanks. I actually wonder if it would be neater to just reference `scipy`... as long as you don't prevent pynbody as a whole from importing if scipy is not present. So either import scipy within the one function, or at file level but with a suitable exception handler?

Scipy is so ubiquitous these days that it is tempting just to make it a dependency, but this half-way-house could be good for now. Let me know what you think.

biernackip:
What I wondered is if it's not easier to implement the functions that convert the birth times straight in pynbody... I will try and see if I can manage to implement that. It would eliminate the need of calling `part2birth`, reading extra files, importing `fortranfile` etc.

apontzen:
Yes if you have time to do that it would be awesome

biernackip:
I tried the direct approach, but it is very slow, as one needs to integrated the time for each star. RAMSES is using a tabulated approximate values, but when I tried this in pynbody it was also very slow. 

I think the current solution is most efficient, if complex. I can convert this PR to use scipy rather than extra file.

What do you think?

apontzen:
Worst case, one could always build a cython wrapper around the fortran code - but if it is using a look-up table there should be fast ways to implement that in pure python (especially with access to scipy).

So my preference would be to find a good implementation that can be used directly! But if you don't have time or energy for this, I'd go for the original option of pointing directly to scipy (just making sure it doesn't break using other parts of pynbody without scipy).

biernackip:
My first idea was to duplicate the fortran solution. There the values were just approximate (originally size of the table was 1000) and then interpolated. This requires finding the right index for each star. Fortran just does it in a loop over stars, but that was slow too. Is there a better way than looping over stars in pynbody? (sorry, don't know cython)

This whole thing should become obsolete, there is a way of forcing RAMSES to output proper time, just not default (I'll be rooting to change that)

apontzen:
Thanks!

PR #446, opened 2018-03-20
Adding ability to calculate the halo mass function from simulation data

Martin-Rey:
Pynbody currently has the ability to generate theoretical fits to the halo mass function. 

This PR adds a function to calculate the halo mass function by binning halo mass from the snapshot catalogue. It allows for quick comparison between the two.

Comments more than welcome.

I also have a short notebook with a recipe to compare both outputs. Let me know if this is worth sharing and what is the best way to do it.


apontzen:
Ideally the recipe would be added in the `docs/tutorials` area (http://pynbody.github.io/pynbody/tutorials/tutorials.html)

Otherwise this looks ready to merge...

Martin-Rey:
I am almost there with writing a new tutorial. 

Quick question, in the test data, is there a simulation of uniform resolution with a halo catalogue to generate a HMF plot ? I found a tiny zoom simulation but obviously the HMF generated from it is wrong. 

This also showed that this implementation will not handle zooms. I'll add a quick assert on the mass array to make sure that nobody is tricked by this.

apontzen:
There isn't a uniform volume simulation in the pynbody test data but you could maybe provide a link to the tangos data?

Re zooms I would make it a warning rather than an exception. You never know, someone might intentionally want to do this.

Martin-Rey:
Don't you think It would be neater to keep pynbody and tangos separated for a tutorial ? Otherwise, yes I can add a database and obtain halo masses from there.

apontzen:
I didn’t mean to use any tangos code - just point to one of the sample files (the gadget one should do) to get a uniform volume run.

apontzen:
Thanks - one last request though, could you explain where people should obtain the data from in `halo.rst`

Martin-Rey:
I added an html to the tangos tutorial to get the simulation data. Is that what you had in mind ?

PR #451, opened 2018-05-09
Adding ability to generate profile functional form

Martin-Rey:
This PR adds a new analysis tool to complement the existing profile tools. It is a framework to generate theoretical functional form for profiles. As an example, the given implementation is a NFW density profile with its different parameters.

The abstract base class implements a basic fitting routine based on standard leastsquare. 

This can serve as a base to generate further profile functional form, for density or any quantity as a function of radius.

All comments welcome.


apontzen:
Thanks!

PR #456, opened 2018-05-31
Halo shape with radius

Morkney:
Hello!

Added here is a function that will fit homeoidal shells to a DM halo, based on the iterative calculation of the 'shape tensor'. 

The function has a rate limiting step in the handling of the particle data, which can make the process quite slow when handling a large number of particles. 

Any comments on how to improve the code, or adapt the style to fit better with Pynbody, are welcome!

### DIFF
--- a/pynbody/analysis/halo.py
+++ b/pynbody/analysis/halo.py
@@ -311,3 +311,126 @@ def center(sim, mode=None, retcen=False, vel=True, cen_size="1 kpc", move_all=Tr
         tx = transformation.inverse_v_translate(tx, velc)
 
     return tx
+
+def halo_shape(sim, N=100, rin=0, rout=0, bins='equal'):
+
+    """
+
+    Return the axis ratios b/a and c/a, and the alignment angle, for homeoidal shells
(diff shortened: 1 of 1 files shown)


apontzen:
Hi @morkney, it looks like a nice addition to me.

Could you modify the docstring slightly so that the arguments are laid out a bit more clearly/explicitly. The docstrings in pynbody are a bit inconsistent but one reasonable example is here:

https://github.com/pynbody/pynbody/blob/master/pynbody/plot/sph.py#L286

Unlike in that example, there should not be any new lines at the top of the docstring either.

Otherwise I am happy to merge this. Re performance - I can't immediately see an optimization so will leave it with you to see if any can be found.

Morkney:
Thank you for your advice @apontzen! 

I have made some changes to the docstring to better match those in Pynbody. I will continue to look at optimization.

apontzen:
@morkney I see you are still tinkering - are you happy for me to merge this now or would you like longer?

Morkney:
@apontzen Yes, I think I'll leave it as it is now.

apontzen:
Thanks for this nice addition!

PR #458, opened 2018-06-07
Ability to define halo boundary with respect to critical density

Martin-Rey:
Currently, the halo boundary can be calculated through the virial_radius() command. It is possible to define different overdensity threshold (178, M200, M500 etc) but not to choose whether the overdensity is calculated against the matter or the critical density.

This PR adds a "rho_def" keyword to the method to allow for this.

Any comments welcome !


apontzen:
Could you update the docstring to explain the possibilities for rho_def? Otherwise looks good!

Martin-Rey:
I am unsure why the last test has failed. It exceeded the Travis time limit of 50min but at compile time. I don't think it has anything to do with this PR, especially since 3 out of 4 tests have worked.

Is it possible to restart the build ?

Issue #355, opened 2017-01-23
issue with v_circ calculation


astroduff:
Hey gang, happy new year! Noticed an issue with calculating v_circ for profile when using the HDF5 Gadget3. Essentially the gravity calc used for estimating v_circ is calling 

"import pynbody.gravity.calc as gravity" 
rather than 
"import pynbody.analysis.gravity as gravity"
the latter is very slow (but works), the former results in a type error report, however which reports a read type error (as below) which is bizarre because having checked the types being passed none are 'float', see below...

/Users/aduffy/anaconda/lib/python2.7/site-packages/pynbody/gravity/calc.pyc in midplane_rot_curve(f, rxy_points, eps, mode)
    106         fn = mode
    107 
--> 108     pot, accel = fn(f, np.array(rs), eps=np.min(eps))
    109 
    110     u_out = (accel.units * f['pos'].units) ** (1, 2)

pynbody/gravity/_gravity.pyx in pynbody.gravity._gravity.direct (pynbody/gravity/_gravity.c:1819)()

ValueError: Buffer dtype mismatch, expected 'DTYPE_t' but got 'float'

 

Thoughts? I can pass HDF5 files although they're a little on the large size.

apontzen:
Thanks - I guess the problem is that direct.pyx seems to be expecting double precision whereas your file is in single precision.

This should be solvable by using cython's fused types (like _util.pyx does)...

astroduff:
Hey Andrews

Yes, unfortunately the fused type:
ctypedef fused DTYPE_t:
    np.float32_t
    np.float64_t 

then breaks the following call from math.h

cdef extern from "math.h" nogil :
      DTYPE_t sqrt(DTYPE_t)

I’m afraid this is all above my head, but just for fun I set the DTYPE_t to be np.float32_t and got the following

/Users/aduffy/anaconda/lib/python2.7/site-packages/pynbody/gravity/calc.pyc in midplane_rot_curve(f, rxy_points, eps, mode)
    106         fn = mode
    107 
--> 108     pot, accel = fn(f, np.array(rs), eps=np.min(eps))
    109 
    110     u_out = (accel.units * f['pos'].units) ** (1, 2)

pynbody/gravity/_gravity.pyx in pynbody.gravity._gravity.direct (pynbody/gravity/_gravity.c:1587)()

ValueError: Buffer dtype mismatch, expected 'DTYPE_t' but got 'double'



[...]

PR #461, opened 2018-06-18
Introduce a common function to retrieve the `eps` quantity

elehcim:
Tidy up `eps` usage.
I added the function `get_eps` in util.py which tries to retrieve `eps` taking care of the correct dtype which sometimes could create problems (e.g. dtype mismatch like the one in #355).

### DIFF
--- a/pynbody/gravity/calc.py
+++ b/pynbody/gravity/calc.py
@@ -1,12 +1,15 @@
 from .. import units
 from .. import array
 from .. import config
+from ..util import get_eps
 
 import math
 import tree
 import numpy as np
 
 import warnings
+
+from ._gravity import direct
 
 
 def all_direct(f, eps=None):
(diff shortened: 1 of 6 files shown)


apontzen:
Thanks, this looks great as soon if it can be made compatible with python <3.5 (unclear to me why the travis tests are failing for those python versions at a glance)

pynbody/gravity/calc.py
```diff
@@ -83,14 +83,7 @@ def midplane_rot_curve(f, rxy_points, eps=None, mode=config['gravity_calculation
             "OpenMP module is now selected at install time", DeprecationWarning)
 
     if eps is None:
-        try:
-            eps = f['eps']
-        except KeyError:
-            eps = np.zeros(len(f))
-            eps += f.properties['eps']
-
-    if isinstance(eps, str):
```
apontzen: For backwards compatibility I think we need to allow eps to be a string that gets auto-converted into a unit. This doesn't seem to be possible in the new version.

elehcim: The prerequisite of the function `direct` in _gravity.pyx is that eps should be a SimArray of the same dtype as `f['mass']` and the same units as `f['pos']`. For that before every call to `direct` I think it is good to make sure `eps` is a SimArray.

apontzen:
Many thanks for this excellent patch!

PR #486, opened 2018-09-17
Corrected halo_shape() binning with equal number of particles per shell

b-fontana:
I noticed two possible issues in the pynbody.analysis.halo.halo_shape() method, when one selects the bins='equal' option:
1. The code crashes because at some point an array is being sliced with floats;
2. The implemented binning seems to be just a linear one, contrary to what is described in the documentation.

I am providing a solution for the two above points. Please comment, and correct me if I missed something.

### DIFF
--- a/pynbody/analysis/halo.py
+++ b/pynbody/analysis/halo.py
@@ -398,8 +398,6 @@ def halo_shape(sim, N=100, rin=None, rout=None, bins='equal'):
         mid = sn(np.sort(posr[np.where((posr >= rin) & (posr <= rout))]),N*2)
         rbin = mid[1:N*2+1:2]
         mid = mid[0:N*2+1:2]
-        print(rbin)
-        print(mid)
         
     elif (bins == 'log'): # Bins are logarithmically spaced
         mid = profile.Profile(sim.dm, type='log', ndim=3, min=rin, max=rout, nbins=N+1)['rbins']
--- a/pynbody/analysis/halo.py
+++ b/pynbody/analysis/halo.py
@@ -372,11 +372,25 @@ def halo_shape(sim, N=100, rin=None, rout=None, bins='equal'):
 
     # Define moment of inertia tensor:
     MoI = lambda r,m: np.array([[np.sum(m*r[:,i]*r[:,j]) for j in range(3)]\
-                               for i in range(3)])
+                                for i in range(3)])
 
     # Splits data into number of steps N:
     sn = lambda r,N: np.append([r[i*int(len(r)/N):(1+i)*int(len(r)/N)][0]\
                                for i in range(N)],r[-1])
+
+    def EqualBins(sim, rin, rout, nshells=N, r='down'):
+        """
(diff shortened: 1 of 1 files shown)


apontzen:
Thanks @b-fontana. 

@Morkney would you like to comment on these changes since you implemented the original?

It would be good to implement a nosetest using one of the testdata files so we can keep track of any problems with this function.

As a minor comment, functions should be `written_with_underscores` rather than `WrittenWithCase` ala [PEP8](https://www.python.org/dev/peps/pep-0008/).

Morkney:
Hello,

With regards to the first issue, I've not experienced this myself and it must be due to different versions of Python? Thanks for raising this issue because I would likely never have noticed otherwise.

With the second point, I believe the current code does achieve what is laid out in the documentation. The current code splits the particle distribution into radial shells containing equal numbers of particles. It returns the radii of the shell borders in 'mid' and the radii of the shell centres in 'rbin'. If you were to ask Pynbody:

`len(np.where( (halo['r'] > mid[i]) & (halo['r'] <= mid[i+1]) )[0])`

...where 'halo' is the halo dataset, then the number of particles should be equal for all values of 'i'. 
The array returned by your 'EqualBins' function achieves the same, so I think the two methods are equivalent. If i'm mistaken here then feel free to correct me.

On an unrelated note I've noticed a problem with the way the code handles rotations. In almost all situations this results in no significant change and has therefore slipped under the radar until now. I'll be making a new pull request with the fix.

Best, -Matt

b-fontana:
Perhaps I am missing some detail, so please correct me if I am wrong.
The 'equal_bins' options currently does the following:

`if (bins == 'equal'): 
        mid = sn(np.sort(posr[np.where((posr >= rin) & (posr <= rout))]),N*2)
        rbin = mid[1:N*2+1:2]
        mid = mid[0:N*2+1:2]`

where posr is defined by

`posr = np.array(sim.dm['r'])[np.where(sim.dm['r'] < rout)]`

The shells seem to be exclusively defined by the halo radius (`halo['r']`). I do not see where the connection to the number of particles is done. My suggestion considers instead the number of particles, by using `len(halo)`.

According to @Morkney, the `mid` array already contains equal number of particles in each of its stored distances. If this is true, I did not find the place where this is done.

b-fontana:
Hello again, @Morkney and @apontzen ,

I have now realized I had not undertood the `sn()` function. It is indeed doing what is stated in the documentation, and it does that in a more compact way than my suggestion.
If I am not wrong, the last bin will have a different number of particles than all the other ones. You could maybe consider adding some comment in the code referring that, as well as explaining in more detail what the `sn()` function does.

Regarding the 'slicing with floats' issue, it was corrected implemented.

*Summary:*

1. The slicing is now being correctly done;
2. The 'equal' option is correct; sorry for the misunderstanding.

Cheers.

Morkney:
Hi @b-fontana,
Thank you for your advice. I agree with your suggestions and I'll be tweaking the comments accordingly.

PR #487, opened 2018-09-24
Halo shape with radius (Rotations fix)

Morkney:
Hello!

This is the same halo_shape function as before, but with some changes to the way rotations are handled. This has improved the accuracy of the rotation matrices returned by the function, which would previously be completely incorrect on rare occasions.

Also included is a minor fix to prevent a crash when an array is sliced with floats, as suggested by @b-fontana.

EDIT 1: seems I was hasty, there may still need to be some checks done on this.
EDIT 2: Checks completed and it looks like the problem was fixed. I'm not certain of the rotation matrix convention so they may be transposed the wrong way, but they should be correct in either case.
EDIT 3: I am currently updating the code a little more, including with suggestions from @b-fontana.


apontzen:
I am going to hold off on this until we clarify whether it addresses all the issues in PR #486. Unfortunately we may end up needing a three-way merge.

Morkney:
@apontzen I think everything's finished here now, unless @b-fontana has anything else to add.

b-fontana:
@apontzen, @Morkney : 

I have nothing to add. Feel free to close this PR.

Issue #485, opened 2018-09-11
`IndexedSimArray` problem


elehcim:
I tried to run the script [examples/vdisp.py](../tree/master/examples/vdisp.py) with pynbody 0.46 and python 3.6. I think it should work even if it's an old one. I tried with many Gadget2 datasets including `testdata/test_g2_snap.1`, the error persists.
I have the following problem in accessing an `IndexedSimArray`.

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-36-23f851898348> in <module>()
     13     bininds = np.where(rxyinds == ind)
---> 14     sigvz[i] = np.std(stars[bininds]['vz'].in_units('km s^-1'))
     15 
     16     try:

~/Software/miniconda3/lib/python3.6/site-packages/numpy/core/fromnumeric.py in std(a, axis, dtype, out, ddof, keepdims)
   3022             pass
   3023         else:
-> 3024             return std(axis=axis, dtype=dtype, out=out, ddof=ddof, **kwargs)
   3025 
   3026     return _methods._std(a, axis=axis, dtype=dtype, out=out, ddof=ddof,

~/Software/miniconda3/lib/python3.6/site-packages/pynbody-0.46-py3.6-linux-x86_64.egg/pynbody/array.py in q(s, *y, **kw)
    945         # but it seems required on McMaster setup (Py 2.6.5, NP 1.4.1)
    946         if kw != {}:
--> 947             return w(SimArray(s), *y, **kw)
    948         else:
    949             return w(SimArray(s), *y)

~/Software/miniconda3/lib/python3.6/site-packages/pynbody-0.46-py3.6-linux-x86_64.egg/pynbody/array.py in __new__(subtype, data, units, sim, **kwargs)
    194     def __new__(subtype, data, units=None, sim=None, **kwargs):
    195         new = np.array(data, **kwargs).view(subtype)
--> 196         if hasattr(data, 'units') and hasattr(data, 'sim') and units is None and sim is None:
    197             units = data.units
    198             sim = data.sim

~/Software/miniconda3/lib/python3.6/site-packages/pynbody-0.46-py3.6-linux-x86_64.egg/pynbody/array.py in sim(self)
    906     @property
    907     def sim(self):
--> 908         return self.base.sim[self._ptr]
    909 
    910     @sim.setter

TypeError: 'NoneType' object is not subscriptable
```

[...]

PR #490, opened 2018-09-27
Fix for #485

elehcim:
When using the `in_units` method of an `IndexedSimArray`, a new instance of `IndexedSimArray` is created. After that, for some reason the `SimArray.__del__` method is called and, due to how this is coded [here](https://github.com/pynbody/pynbody/blob/3fac16d1a2fe8170e73828613378021c4726a7ea/pynbody/array.py#L653), it calls in turn the [`SimArray.__new__`](https://github.com/pynbody/pynbody/blob/3fac16d1a2fe8170e73828613378021c4726a7ea/pynbody/array.py#L193) method. This last call produces the issue #485 in python3.

~~I think we don't have to wrap the `SimArray.__del__` method around `IndexedSimArray`~~

As described [here](https://github.com/pynbody/pynbody/issues/485#issuecomment-425438948) I removed the exception thown in `IndexedSimArray.sim` property

### DIFF
--- a/nose/array_test.py
+++ b/nose/array_test.py
@@ -160,8 +160,8 @@ def test_issue_485_1():
     s = pynbody.load("testdata/test_g2_snap.1")
     stars = s.s
     indexed_arr = stars[1,2]
-    np.testing.assert_almost_equal(np.sum(indexed_arr['vz'].in_units('km s^-1')), -38.2072944641)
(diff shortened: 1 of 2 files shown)


apontzen:
Thanks for tracking this down!

But your commit seems to have a test without an actual fix. (Although weirdly the tests are passing...)

elehcim:
Actually I just realized that what I wrote is not exactly true. The problem is when we do something on the IndexedSimArray which has prevously gone through a unit changed. More work needed

PR #494, opened 2018-10-18
Allow disabling cosmological conversions in GadgetSnap at load time

elehcim:
It is useful for some Gadget snapshots I have (e.g. non cosmological simulations but with some header info about the cosmology) to ignore cosmological conversions.

### DIFF
--- a/nose/gadget_test.py
+++ b/nose/gadget_test.py
@@ -191,10 +191,17 @@ def test_issue321():
     assert f['pos'].dtype==np.dtype('float32')
     assert f['mass'].dtype==np.dtype('float32')
 
+
+def test_units_override():
+    f = pynbody.load("testdata/test_g2_snap.1")
+    assert_equals(f['pos'].units, "kpc a h^-1")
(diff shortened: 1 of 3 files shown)


apontzen:
This looks like a useful feature, but is there a reason this would be a configuration option rather than something passed to `load()`? You’d think people are more likely to know at load-time, rather than want to change it globally?

elehcim:
Yes, passing it at loading time is a good idea.

elehcim:
I'd like to test this functionality but the gadget2 snap in `testdata` has a corresponding .units file which reintroduces cosmological factors.
I was thinking some renaming or temp file hacking. But maybe it is simpler adding another copy of gadget data without a .units file.

elehcim:
I understood better how the units override system works, I fixed the relative test and I could introduce a test on the `ignore_cosmology` functionality. Now tests should succeed.

elehcim:
Any opinion against this?

apontzen:
Looks good to me, got lost in the inbox. Thanks.

elehcim:
No problem, thanks.

PR #504, opened 2019-02-13
fix to ahf.py in function _can_load() to take ahf_basename into account when passed to AHFCatalogue

TobiBu:


pynbody/halo/ahf.py
```diff
@@ -52,7 +52,11 @@ def __init__(self, sim, make_grp=None, get_all_parts=None, use_iord=None, ahf_ba
         """
 
         import os.path
-        if not self._can_load(sim):
+        if ahf_basename is not None:
+            can_load_ahf = self._can_load(sim, ahf_basename)
+        else:
+            can_load_ahf = self._can_load(sim)
+        if not can_load_ahf:
```
apontzen: Since `_can_load` already checks for `None`, can you not simplify these 5 lines to `if not self._can_load(sim, ahf_basename)`?

TobiBu: indeed. That was stupid double checking.

TobiBu:
I fixed that now and pushed a new version.

PR #508, opened 2019-03-07
Using custom CMDs

apontzen:
…ript for generating such a file

### DIFF
--- a/nose/tipsy_test.py
+++ b/nose/tipsy_test.py
@@ -2,6 +2,7 @@ import pynbody
 import numpy as np
 import glob
 import os
+from nose.tools import assert_raises
 
 
 def setup():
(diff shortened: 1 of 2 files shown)


elehcim:
Hello Andrew, thanks for this, especially the script in the comment will be quite useful for me. One thing: why not passing the `cmdlum_file` path directly to `calc_mags` function (with default for example to `os.path.join(os.path.dirname(__file__), "cmdlum.npz")`)? I think it would avoid globals and make it explicit which file you are using. Also one doesn't need to remember to call a function to revert it back.

apontzen:
Thanks... the main reason for this is so that people can use the ‘derived arrays’ which automatically calculate the magnitudes. However I agree it would be useful to have also an option to pass the path if directly calling the function. I’ll add this before merging.

apontzen:
@elehcim would be interested to check this update addresses your question

elehcim:
> Thanks... the main reason for this is so that people can use the ‘derived arrays’ which automatically calculate the magnitudes. However I agree it would be useful to have also an option to pass the path if directly calling the function. I’ll add this before merging.

Oh, I see why you had to use globals. Thanks for the modifications.

Issue #520, opened 2019-05-03
Ramses tform utility


Martin-Rey:
Hi,

I have come across two bugs while using the Ramses utility get_tform.

1) If the path to the snapshot has a "/" at the end (following an autocomplete with tab for example), the utility would fail due to its parsing mechanics here:
https://github.com/pynbody/pynbody/blob/cb18c8d0ba847975046db550ec281d3c3cd48e25/pynbody/analysis/ramses_util.py#L437-L444

in which the "/" is taken out in the parent_dir variable, but the top.filename variable is used to extract the 12 characters mapping to the output name. You can see an example following, for which the error follows from this try/catch block:
https://github.com/pynbody/pynbody/blob/cb18c8d0ba847975046db550ec281d3c3cd48e25/pynbody/analysis/ramses_util.py#L455-L464

```
In [1]: s = pynbody.load("./Halo1459_fiducial_fblim/output_00032/")
/Users/mrey/anaconda3/lib/python3.6/site-packages/pynbody-0.47-py3.6-macosx-10.7-x86_64.egg/pynbody/snapshot/ramses.py:731: UserWarning: Using field 'tform' to distinguish stars. If this is wrong, try editing your config.ini, section [ramses], entry particle-distinguisher.
  warnings.warn("Using field %s to distinguish stars. If this is wrong, try editing your config.ini, section [ramses], entry particle-distinguisher."%pb_name)

In [2]: pynbody.analysis.ramses_util.get_t(s)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-2-3ac456de44d7> in <module>()
----> 1 pynbody.analysis.ramses_util.get_t(s)

AttributeError: module 'pynbody.analysis.ramses_util' has no attribute 'get_t'

In [3]: pynbody.analysis.ramses_util.get_tform(s)
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
~/anaconda3/lib/python3.6/site-packages/pynbody-0.47-py3.6-macosx-10.7-x86_64.egg/pynbody/analysis/ramses_util.py in get_tform(sim, part2birth_path, cpu_range)
    456             birth_file = FortranFile('%s/output_%s/birth_%s.out%05d' %
--> 457                                      (parent_dir, top._timestep_id, top._timestep_id, i + 1))
    458         except IOError:

[...]

Issue #231, opened 2015-01-15
For RAMSES runs, luminosity.py has bug on line 53: need "metal" vs "metals"


earnric:
I had to change 

    metals = simstars['metals']
to
    metals = simstars['metal']

I noticed that other routines copy s.s['metals'] = s.s['metal'] so that either work. However, I don't think this happens when working with a RAMSES snap.

I've made the change and can upload after you verify this is the case... Or, since its so simple, I'm sure you can make the change.

Aloha!

Rick S
ASU

apontzen:
Actually I think the error is on line 134 of default_config.ini which names it 'metal'. For every other snapshot type it's called 'metals' so it ought to be for ramses too. Trouble is, by fixing this now we could break some people's code... Hmmm.... tricky

rokroskar:
I think it was done this way to keep in line with the naming convention from RAMSES... so I'm not sure we should change anything in this case

PR #521, opened 2019-05-03
Issue 520

Martin-Rey:
Hi again,

Here is a first shot at resolving the problems highlighted in #520. 

Specifically, this PR addresses 1) path mishandling of ramses tform by simplifying its logic and 3) deleting redundant HOP utilities. 

For 2), I need to think about it a bit more. The current routine replaces the 'tform' array in place, as it is also the name of the simulation array that is read from the actual ramses file (specified in the ramses header). Having it as a derived array would require some name mapping and would be very different in spirit to existing derived arrays.

Martin

### DIFF
--- a/nose/ramses_test.py
+++ b/nose/ramses_test.py
@@ -147,3 +147,14 @@ def test_forcegas_dmo():
 
 def test_metals_field_correctly_copied_from_metal():
     np.testing.assert_allclose(f.st['metals'][::5000], f.st['metal'][::5000], rtol=1e-5)
+
+
+def test_tform_and_tform_raw():
+    # Tform is not available for test data, so test warning generation and that the array is full of -1
+    with np.testing.assert_warns(UserWarning):
+        assert len(f.st['tform']) == len(f.st['tform_raw']) == 2655
+        np.testing.assert_allclose(f.st['tform'], - np.ones((2655,), dtype=np.float64), rtol=1e-5)
+
+    np.testing.assert_allclose(f.st['tform_raw'][:10], [4.58574701, 4.58100771, 4.58284129, 3.18777836, 4.55801122,
+                                                        4.50733498, 4.5100136,  4.57288808, 4.55926183, 4.52128465],
+                               rtol=1e-5)
--- a/nose/ramses_test.py
+++ b/nose/ramses_test.py
@@ -143,3 +143,7 @@ def test_forcegas_dmo():
             2.25179981e+15,   2.25179981e+15,   2.25179981e+15,
             2.25179981e+15,   2.25179981e+15,   2.25179981e+15,
             1.80143985e+16], rtol=1e-5)
+
+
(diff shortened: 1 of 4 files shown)


apontzen:
As in #520 - I changed my mind and don't think it's actually possible to make it a derived array at a technical level. But I do think it's correct to rename the existing `tform` to something else, since it is not in any sense the same as `tform` that gets stored by tipsy or gadget files.

Martin-Rey:
Alright, so here is an implementation hiding the ramses 'tform' native field into a 'tform_raw' array and reading 'tform' from the birth files. I have left most of the implementation inside the analysis/ramses_util directory, but this can be moved around easily.

In the best of world, there should be test for the new tform implementation, but none of the test dataset have birth files, making it currently impossible. Would you be ok with me generating and uploading these files on one ramses dataset to the testadata?

The last question for me is dealing with the exceptions generated by the ramses utility. What should be the behaviour of `snapshot.star['tform']` when 1) no birth files are found in the directory and 2) the ramses utility part2birth is not found. Should it print a warning and return the Ramses default value? Should it throw an exception with a message explaining that the default Ramses field can be found at `'tform_raw`?

I used this opportunity to also generate a mapping for the ramses 'metal' field, into a 'metals' field, to solve the recurrent issue of having to copy this field to use analysis routine such as luminosity calculations (see e.g. #231). The array is copied to be backward compatible for people using 'metal'. Let me know what you think about this addition.

I also removed the derived `'tform` array at the bottom of the Ramses file, as I don't think there is a case in which we do not want to read it from disk. Let me know if I am wrong in this.

Best,
Martin

Martin-Rey:
Hi @apontzen,
 
I have added tests of the new loader functionalities and error handling. This branch is almost ready to merge.

The test are currently failing due to an inconsistent default name for the formation time array when it can't be read from the header as is the case for the testdata:
1) Ramses (old format but latest stable branch) by default calls it 'tform', as expected in the pynbody utilities.
2) Pynbody default configuration calls it 'age', which is usually a derived quantity from 'tform' throughout pynbody. 
https://github.com/pynbody/pynbody/blob/b37d0dba99d7b64ab2fc51355d8068afb609d7ab/pynbody/default_config.ini#L185
This seems to suggest that the ramses field has meant different things over the years?

I propose to update the default config, replacing 'age' by 'tform', but won't do it before I have your opinion on whether this would break other people's code.

Martin

Martin-Rey:
Alright, that should be good to go. As we discussed, I modified the ramses default 'age' to 'tform'.

Sorry for the multiplication of commits, I had forgotten that python2 and 3 have different behaviours for IOError and OSErrors.

Martin

Martin-Rey:
Hi,

Just a quick check whether there is something holding this PR from being merged in the master branch?

Martin

apontzen:
Don't think so! Thanks for the reminder.

PR #534, opened 2019-08-13
ramses read cpus and namelist

mkrets:
This fixes `_load_particle_cpuid` for ramses.
This introduces also `_namelist` similar to `_info` to quickly have access to the namelist parameters.


mkrets:
The namelist in testdata/ramses_dmo_partial_output_00051 seems to be broken.

apontzen:
Weird, I agree it doesn't look right! Shouldn't that mean the unit tests fail now that you are loading it?

mkrets:
Initially the unit tests failed but I avoided this by adding a try/except statement.
I am not sure if this is

```python
try:
    self._load_namelist_from_specified_file(f)
except ValueError:
    warnings.warn("Namelist found but unable to read.")    
```

apontzen:
OK, fair enough! As a final touch is it possible to add a unit test that verifies your namelist reading is working (using e.g. `ramses_new_format_partial_output_00001`)?

mkrets:
I added some simple unit tests - maybe we want something more thorough?

I noticed that some values (e.g. `1e-10`) are not transformed to float.
Therefore this part can be improved:
```python
if '.' in val:
    self._namelist[name] = float(val)
...
```

(same in `_load_info_from_specified_file`).

mkrets:
Although this PR was merged, the changes are not in there anymore.
@apontzen could you have a look if something went wrong with the merging?

apontzen:
How bizarre! I will see if I can figure out what has happened.

PR #538, opened 2019-08-31
Fix issue 537

TobiBu:
made a small fix to ahf.py in order to be able to deal with AHF_particle files which have more than two columns.


apontzen:
There's a lot of ancillary changes and extra files committed here - could you strip it back to a simpler patch?

TobiBu:
oh, damn, I thought I added just ahf.py...
sorry.

apontzen:
Easy mistake to make ;-)

PR #539, opened 2019-08-31
Fix issue 537

TobiBu:
fix to _load_particle_block to comply with AHF_particles files which have more than two columns.


apontzen:
Thanks and sorry for the delay in responding to this PR

Could you change the variable `special` to something more descriptive (and also protected), like `_multi_column_particle_file`? Otherwise looks good!

PR #546, opened 2019-12-10
AdaptaHOP support

cphyc:
This adds support for the AdaptaHOP halo finder.

TODO:
- [x] Add data to the testsuite. Could I get some insight for this?

Related to #548 .


cphyc:
The test files I used have a size of 222Mio. If we want to include them in the test suite, they will need to be added to http://star.ucl.ac.uk/~app/testdata.tar.gz.

nose/adaptahop_test.py
```diff
@@ -0,0 +1,15 @@
+import pynbody
+from pynbody.halo.adaptahop import AdaptaHOPCatalogue
+
+import numpy as np
+
+
+def test_load_adaptahop_catalogue():
+    f = pynbody.load('/home/ccc/tmp/output_00099')
```
apontzen: I've updated the testdata on the server but this points to a hard-coded path

cphyc:
This should be merged once #548 is merged.

PR #551, opened 2020-01-17
Adaptahop support improvement

cphyc:
This PR improves the support of adaptahop in pynbody (see #546 for the original PR).

It changes the following behaviour:
- the halo data is not cached any more, as it was causing memory issues,
- to mitigate the computational cost of the aforementioned change, I added an OpenMP-parallelized binary search algorithm that finds items of a sorted array into another sorted array. This improves by a factor of ~2xNCPU the performance compared to the `np.searchsorted` function. This was used to generate an `IndexedSubSnap` from a list of particle ids,
- a new derived array has been defined, namely `iord_argsort` which is the result of `np.argsort(sim['iord'])`,
- implement `get_group_array` for adaptahop, to make it usable with https://github.com/pynbody/tangos (see https://github.com/pynbody/tangos/issues/112).

### DIFF
--- a/nose/adaptahop_test.py
+++ b/nose/adaptahop_test.py
@@ -1,8 +1,7 @@
-import pynbody
-from pynbody.halo.adaptahop import AdaptaHOPCatalogue
-
-from scipy.io import FortranFile as FF
 import numpy as np
+from scipy.io import FortranFile as FF
+
+import pynbody
 
 
 def test_load_adaptahop_catalogue():
@@ -12,9 +11,17 @@ def test_load_adaptahop_catalogue():
 
 
 def test_load_one_halo():
-    f = pynbody.load("testdata/output_00080")
-    h = f.halos()
-    np.testing.assert_allclose(h[1].properties["members"], h[1]["iord"])
+    def helper(path, nhalos):
+        f = pynbody.load(path)
+        h = f.halos()
+        np.testing.assert_allclose(h[1].properties["members"], h[1]["iord"])
+        assert len(h) == nhalos
+
+    for path, nhalos in (
+        ("testdata/output_00080", 170),
+        ("testdata/new_adaptahop_output_00080", 2),
+    ):
+        yield helper, path, nhalos
 
 
 def test_get_group():
--- a/nose/adaptahop_test.py
+++ b/nose/adaptahop_test.py
@@ -6,43 +6,45 @@ import numpy as np
 
 
 def test_load_adaptahop_catalogue():
-    f = pynbody.load('testdata/output_00080')
+    f = pynbody.load("testdata/output_00080")
     h = f.halos()
-    assert len(h) == h._headers['nhalos'] + h._headers['nsubs']
+    assert len(h) == h._headers["nhalos"] + h._headers["nsubs"]
+
 
 def test_load_one_halo():
-    f = pynbody.load('testdata/output_00080')
+    f = pynbody.load("testdata/output_00080")
     h = f.halos()
-    np.testing.assert_allclose(h[1].properties['members'], h[1]['iord'])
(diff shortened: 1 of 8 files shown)


pynbody/snapshot/__init__.py
```diff
@@ -1908,13 +1908,14 @@ def __init__(self, base, index_array=None, iord_array=None):
     def _iord_to_index(self, iord):
         # Maps iord to indices. Note that this requires to perform an argsort (O(N log N) operations)
         # and a binary search (O(M log N) operations) with M = len(iord) and N = len(self.base).
-        if 'iord_argsort' not in self.base:
-            self.base['iord_argsort'] = np.argsort(self.base['iord'])
+
+        if not util.is_sorted(iord) == 1:
+            raise Exception('Expected iord to be sorted in increasing order.')
```
apontzen: As far as I can tell, it seems like it's still ok if iord isn't sorted, since `iord_argsort` takes care of it? But maybe I missed something?

apontzen:
Looks good! I just had one query flagged inline...

pynbody/snapshot/__init__.py
cphyc: Unfortunately no, because `util.binary_search` expects its first argument to be sorted in increasing order (unlike `np.search_sorted`).

cphyc: Maybe I made a poor choice of variable name ; `iord_argsort` is an array that sorts `iord_base`, not `iord`. I am changing the variable names

nose/adaptahop_test.py
```diff
@@ -12,4 +12,20 @@ def test_load_adaptahop_catalogue():
...
+    iord = f.dm['iord']
+
+    for halo_id in range(1, len(h)+1):
+        mask = group_array == halo_id
+
+        # Check that the indices
+        # - read from halo (h[halo_id]['iord'])
+        # - obtained from get_group_array masking
+        # are the same (in term of sets)
+        assert len(np.setdiff1d(iord[mask], h[halo_id]['iord'])) == 0
```
apontzen: Should you also be checking the reverse? I think `setdiff1d` is asymmetric

pynbody/snapshot/__init__.py
apontzen: Ah I see, yes!

pynbody/halo/adaptahop.py
```diff
@@ -165,11 +178,75 @@ def _read_halo_data(self, halo_id, offset):
...
+
+        igrp = np.zeros(len(data), dtype=int) - 1
+
+        if group_to_indices:
+            grp2indices = {}
+        with FortranFile(self._fname) as fpu:
+            for halo_id, halo in self._halos.items():
+                fpu.seek(halo.properties['file_offset'])
+                fpu.skip(1)  # number of particles
+                particle_ids = fpu.read_vector('i')
```
cphyc, line 237: After #584 is merged, this should be fixed to take into account cases where `particle_ids` is not sorted and/or is not a 32bit integer.

cphyc:
@apontzen do you think this is ready to go?

apontzen:
Is there any way to test this? Happy to add a small file to the testdata tar if needed.

cphyc:
I can generate one of about ~50Mio, would that suffice?

apontzen:
Sounds good, thank you!

cphyc:
We can use the following halo catalog (very minimalistic, only two halos and linked to the `output_00080` dataset already present in the testdata tar).
[tree_bricks080.zip](https://github.com/pynbody/pynbody/files/6436096/tree_bricks080.zip)

apontzen:
Thanks!

Issue #559, opened 2020-04-15
ramses star ages broken


mkrets:
Hi,

in my cosmological ramses runs I use `use_proper_time=.true.` .

I noticed that I had to put back `birth_time: tform` in `.pynbodyrc`.

`data.s["age"].in_units("Gyr")`
gives: UnitsException: Unknown units

but 
`data.s["tform_raw"].in_units("Gyr")`
yields some negative values.

It seems that 
`data.properties["time"].in_units("Gyr") + data.s["tform_raw"].in_units("Gyr")`
gives the correct birth times.

You can find an example output here:
https://www.dropbox.com/s/tb85zmril8p60tx/output_00030.tar.gz

It would be great if `use_proper_time=.true.` would be supported either through `.pynbodyrc` or automatically by reading the namelist (I had something in mind with #534).

I hope that describes the problem well enough.
Thank you for your help.

mkrets:
A quick fix is to modify `_convert_tform`:

```python
def _convert_tform(self):
    # Copy the existing array in weird Ramses format into a hidden raw array
    self.star['tform_raw'] = self.star['tform']
    self.star['tform_raw'].units = self._file_units_system[1]
    if config_parser.getboolean("ramses", "proper_time", fallback=False):
        t0 = analysis.cosmology.age(self, z=0.0, unit="Gyr")
        birth_time = t0 + self.s["tform_raw"].in_units("Gyr")/self.properties["a"]**2
        birth_time[birth_time > t0] = t0 - 1e-7
        self.star['tform'] = birth_time
    else:
        from ..analysis import ramses_util
        # Replace the tform array by its usual meaning using the birth files
        ramses_util.get_tform(self)

```

and then add in the config `.pynbodyrc`:

```
[ramses]
proper_time: True
...
```
If its not present the code above will fall back to False and execute the ramses_util part.

I still need to add `birth_time: tform` as well.
```
[ramses-name-mapping]
# For RAMSES format post November 2017: map the names in the part_file_descriptor.txt to pynbody names
position_x: x
position_y: y
position_z: z
velocity_x: vx
velocity_y: vy
velocity_z: vz
identity: iord
levelp: level
birth_time: tform

...
```

Hope that helps.

YuankangLiu:
> A quick fix is to modify `_convert_tform`:
> 
> ```python
> def _convert_tform(self):
>     # Copy the existing array in weird Ramses format into a hidden raw array
[...]

PR #561, opened 2020-05-13
dealing with star particle ages in ramses when proper time is used

mkrets:
When `use_proper_time=.true.` is used in a cosmological ramses simulation the formation time of star particles is saved in proper time. Namely there is no need to convert them (through `ramses_util.get_tform`).
The original issue is  #559 where I also explain the code-snippet.

To enable  the proper_time handling,  I add in the config .pynbodyrc:

```
[ramses]
proper_time: True
...
```

I had to add an ugly try/except block because python 2.7 doesnt support the fallback option in `config_parser.getboolean`.

I still need to add birth_time: tform as well.

```
[ramses-name-mapping]
position_x: x
position_y: y
...
birth_time: tform

...
```

and for the particle-blocks I also use tform
```
particle-blocks=x,y,z,vx,vy,vz,mass,iord,level,tform,metal
```

Alternatively `use_proper_time=.true.` could be directly read from the namelist (e.g. through #534 )


apontzen:
Thanks! @Martin-Rey could you confirm this looks good to you? If so I am happy to merge.

Martin-Rey:
Hi @mkrets,

Thanks a lot, this looks very good. Could you add the proper time flag in the default pynbody config file of the repository (https://github.com/pynbody/pynbody/blob/0b63e9be4aa34c6577f37bbd18f2261c2b2f0336/pynbody/default_config.ini#L170), with a line explanation such as:
```
# Use proper time when reading the age of star particles. 
# Should be turned to true if e.g. the namelist flag (use_proper_time) or radiative transfer have been used
proper_time: false
``` 

It would probably be good to implement a test catching this behaviour in the future, which should be fairly straightforward. There is an RT example run in the tesdata, so one could imagine turning the new `proper_time` flag on before loading the RT test (https://github.com/pynbody/pynbody/blob/0b63e9be4aa34c6577f37bbd18f2261c2b2f0336/nose/ramses_test.py#L72-L73) and adding an additional verification that the `tform` arrays produced are what we would expect. I am just to sure how you would modify the config file at runtime, any ideas @mkrets  and @apontzen ?

Martin

mkrets:
Thanks for the replies.
I added the flag to the default config.

I agree a test would be good but yes one would have to switch the flag during runtime. I am not sure if that is possible. I looked at the `testdata/ramses_rt_partial_output_00002` but it seems that it does not contain stars.

One thing that confuses me is that I have to add:
```
[ramses-name-mapping]
birth_time: tform
```
Otherwise I get:
```python
ValueError: 'tform' is not in list
```
Should this also be in the default config?

Martin-Rey:
Ah that's annoying to have to add additional data for the test. Let's see what @apontzen prefers and whether it is possible to change such config at runtimes?

Re- birth time: Pynbody tries to guess the name of particle fields from the header file (`output_00XXX/header.txt`) if I recall correctly. Is it possible that your header uses birth_time (which is the internal Ramses name) as the "name" of the `tform` array, hence why you have to remap it? 

If birth_time has become the default name for this array with the new Ramses format, yes it might worth adding as the default config.

apontzen:
There isn't a nice way to do this, but you could probably have the test code change the appropriate variable in `pynbody.snapshot.ramses`. Not pretty, but better than leaving it untested I suspect.

mkrets:
@Martin-Rey I checked and pynbody tries to load the particle field names from the `part_file_descriptor.txt` if that fails pynbody tries to guess it from the `header.txt.`

By default ramses writes `"birth_time"` in `part_file_descriptor.txt` : [see here](https://bitbucket.org/rteyssie/ramses/src/3e75d1ced915d53ca27cbab730282434ca2428b9/pm/output_part.f90#lines-167)

But `'tform'` in `header.txt` : [see here](https://bitbucket.org/rteyssie/ramses/src/3e75d1ced915d53ca27cbab730282434ca2428b9/amr/output_amr.f90#lines-542)

I propose adding `birth_time: tform` to `[ramses-name-mapping]`

Martin-Rey:
@mkrets That makes sense to me. `tform` must be the historical name, that has been replaced by `birth_time` in the latest Ramses iteration. I agree it would be good to have this mapping done by default.

If you can manage to design a test using a small RT (or proper time) snapshot (for example using only 2-3 cpus of a bigger sim), it would be great. To enable testing, we could store the result of detecting proper time into a protected member (e.g. `self._is_using_proper_time`) of the `RamsesSnap` class, rather than the local variable `is_proper_time`. This member can then be accessed (and modified) at runtime, but only for users that know what to look for?

mkrets:
I added a test. A small test output can be found here:

https://www.ics.uzh.ch/~mkretsch/prop_time_output_00030.tar.gz

Martin-Rey:
Many thanks! @apontzen, could you add it to the testdata and re-run the travis build?

mkrets:
I just noticed that ramses writes "metallicity" as the field name for metals in `part_file_descriptor.txt`  ([here](https://bitbucket.org/rteyssie/ramses/src/3e75d1ced915d53ca27cbab730282434ca2428b9/pm/output_part.f90#lines-177)). So additionally I had to add 
```
metallicity: metal
```
to `[ramses-name-mapping]` in the config. Should this also be in the default config?
@apontzen did you have a chance to look at the testdata I uploaded?

apontzen:
Sorry for the delay, I've added this to the tar and restarted the travis build

Martin-Rey:
Thanks @apontzen, looks like the new tests have passed 

@mkrets I think the metallicity mapping should indeed be added to the default config file, but only if `metallicity` is the default field name in the ramses master branch (note that this remapping is now unrelated with your original problem of proper time). I think the default choice should rather be “metals” (with an s), as in
```
metallicity: metals
```

as this is the spelling expected by most analysis routine in pynbody (allowing to calculate luminosities, half-light radius etc), and is the reason for the hardcoded remappings introduced in PR #521 when using the old ramses format for particle fields.

apontzen:
Just wanted to check in with you @mkrets and @Martin-Rey about whether this is ready to merge? As I understand it, there is still some residual question about the default name mappings?

mkrets:
Good to go from my side.
The question wether to add `metallicity: metals`  to the default config is not directly related to this PR.

apontzen:
Thanks to you both!

earnric:
Just a note for those using RAMSES RT sims that store time as proper time.

Make sure you force pynbody to load the tform array (e.g. - len(snap.stars['tform']) ) BEOFRE manipulating the arrays. 

I need to ensure that snap.s returns only active star particles, and not RAMSES debris particles. This may not be the best way to do it, but I simply:

    snap.s = snap.s[snap.s['iord'] > 0]

However, if you do this BEFORE loading tform, this causes a mismatch in the length of the arrays! If you load after this statement the tform array will be LONGER than you other star particle arrays and the values won't match up.

PR #568, opened 2020-07-27
Restore Kravtsov and Hudson SMHMR plotting functions

bwkeller:
It looks like PR #401 removed the Hudson+ 2015 stellar-mass halo-mass relation and duplicated a version of the behroozi SMHMR function that removed the option to use the Kravtsov SMHMR.  This should restore them.


apontzen:
Thanks Ben, but this now seems to remove the Behroozi fits - was that intentional?

bwkeller:
Yep, this is intentional.  If you look at the current version of `stars.py`, there are actually 2 definitions of `behroozi()`.  The second one, which I remove, overwrites the version that includes the extra parameters that let you use the Kravtsov SMHMR.

apontzen:
Ah yes, thanks!

PR #569, opened 2020-07-27
Allow pynbody to infer AREPO HDF5 units

bwkeller:
AREPO has changed where some of the unit information is stored in a Gadget-HDF5 file.  This change should let cosmo/non-cosmo AREPO snapshots to get read with units properly loaded, rather than always falling back to the gadget cosmo units.


apontzen:
Is there any chance of getting a regression test based on a small test file?

bwkeller:
Sure, would it be a good idea just to base it off of the gadgethdf test that currently is in the nose tests?

apontzen:
Sure, that sounds good!

bwkeller:
I've added an extra set of nose tests, just simple checks to make sure the new AREPO units reading is working for both cosmological and non-cosmological zooms.  Since the units are all I really changed, I didn't duplicate the tests that already exist for gadget and gadget HDF snapshots.

bwkeller:
Oh, the tests are failing because the testdata is missing my new test files.  I've put up the ones we need for the tests here: https://project2501.ca/arepo.tar.gz

apontzen:
Thanks @bwkeller - I added the files to the testdata tar but unfortunately the tests still fail (e.g. https://travis-ci.org/github/pynbody/pynbody/jobs/733354558) - now due to a problem with using units in `assert_almost_equal`. Could you check?

PR #576, opened 2020-08-28
Add functionality to infer starlog from log file

emapple:
A recent [changa update](https://github.com/N-BodyShop/changa/commit/125d7e92808744f493cba51b6dc39506601b131d) adds metadata about the starlog file to the output log file. This PR adds functionality to first try to read the log file to determine the starlog structure, and then if that fails continues as before with a guess-and-check approach.

The new `StarLog` method `_infer_name_from_tipsy_log` converts from file types written in the changa output log to the standard pynbody keys.

### DIFF
--- a/pynbody/snapshot/tipsy.py
+++ b/pynbody/snapshot/tipsy.py
@@ -1414,7 +1414,7 @@ class StarLog(SimSnap):
         if meta_name in conversion_dict.keys():
             return conversion_dict[meta_name]
         else:  # this is where cleverness will be needed in the future
-            raise ValueError(f'Unknown starlog entry: {meta_name}')
+            raise ValueError('Unknown starlog entry: ' + meta_name)
 
 
 @TipsySnap.decorator
--- a/pynbody/snapshot/tipsy.py
+++ b/pynbody/snapshot/tipsy.py
@@ -1153,7 +1153,7 @@ def ljeans_turb(self):
 
 class StarLog(SimSnap):
 
-    def __init__(self, filename, sort=True, paramfile=None):
+    def __init__(self, filename, sort=True, paramfile=None, use_log=True):
         import os
         super(StarLog, self).__init__()
         self._filename = filename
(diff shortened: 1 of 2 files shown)


pynbody/snapshot/tipsy.py
```diff
@@ -1181,61 +1169,115 @@ def __init__(self, filename, sort=True, paramfile=None):
...
-        if (iSize > file_structure.itemsize):
+        bigstarlog = False
+        molecH = False
+        bigIOrds = False
+
+
+        if use_log:
+            # assumes log file would be in the same location as starlog file
+            logger.info('Attempting to load starlog metadata from log file')
+            self._logfile = filename.replace('starlog', 'log')
```
apontzen, line 1180: This looks dangerous in principle, since someone might have `starlog` as part of their file or path name

Perhap, ahead of `if use_log`, something like

```python
self._logfile, n_replacements = re.subn(r"\.starlog$", ".log", filename)
if(use_log and n_replacements != 1):
    warnings.warn('Unable to determine log filename; reverting to guess-and-check')
    use_log = False
```

(I didn't test this)

pynbody/snapshot/tipsy.py
```diff
@@ -1181,61 +1169,115 @@ def __init__(self, filename, sort=True, paramfile=None):
...
+                                                       'f8', 'i4')})
+                molecH = False
+
+                if (iSize != file_structure.itemsize and iSize != 104):
+                    raise IOError("Unknown starlog structure iSize:" + \
+                        str(iSize) + ", file_structure itemsize:" + \
+                        str(file_structure.itemsize))
+                else:
+                    bigstarlog = True
+            if molecH == True: print("h2 information found in StarLog!")
```
apontzen, line 1279: This should probably be `logger.info` rather than `print`

apontzen (changes requested):
Thanks -- this looks very helpful. I added a couple of minor comments. 

I also wondered if it would be possible to create some kind of simple test? It is a definite oversight that there is no starlog in the current test suite and it would be nice to add one. There is a `g15784.lr.starlog` in the test data suite which could be used, and one could write a `.log` file temporarily to see if it has the right effect.

pynbody/snapshot/tipsy.py
emapple: While I agree that this belongs in `logger.info` rather than as `print`, this print statement existed before, along with, e.g.,  [this one](https://github.com/pynbody/pynbody/blob/cfbefac64bfad3b8d042e669ff2b36bad25813ce/pynbody/snapshot/tipsy.py#L945). Do you want it addressed here, or in a separate dedicated PR to cleaning up `print`s that should be `logger.info`s?

pynbody/snapshot/tipsy.py
emapple: I have changed it so only the last `starlog` is replaced, which should be safe, given that earlier searches assume the starlog file is the file that ends in `*.starlog`. If you still prefer I use your approach, I'm happy to change it.

I have also added two starlog reading tests - one with and one without a log file. Let me know if this is what you had in mind.

apontzen (approved):

pynbody/snapshot/tipsy.py
apontzen: If you are willing to do a separate PR cleaning up these other instances, that would be much appreciated

PR #582, opened 2020-09-08
Reinstate the work of @mkrets, pull request #534

apontzen:
@mkrets I really can't understand what has happened to your commits from PR #534. The commits are still there but
don't seem to have any effect on the files. 

Probably a git whizz could explain what is going on, but as an immediate practical solution I suggest just re-committing your changes. Could you confirm this pull request captures them correctly?

### DIFF
--- a/nose/ramses_new_ptcl_format_test.py
+++ b/nose/ramses_new_ptcl_format_test.py
@@ -122,3 +122,9 @@ def test_load_no_ptcls():
     assert len(f.dm)==0
     assert len(f.star)==0
     assert len(f.gas) == 196232
+
+def test_loaded_namelist():
+     assert f._namelist['cosmo'] == False
+     assert f._namelist['hydro'] == True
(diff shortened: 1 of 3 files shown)


nose/ramses_new_ptcl_format_test.py
```diff
@@ -121,4 +121,10 @@ def test_load_no_ptcls():
...
     assert len(f.star)==0
-    assert len(f.gas) == 196232
\ No newline at end of file
+    assert len(f.gas) == 196232
+
+def test_loaded_namelist():
+     assert f._namelist['cosmo'] == False
+     assert f._namelist['hydro'] == True
+     assert f._namelist['z_ave'] == 0.1
+     assert f._namelist.get("non_existent_key") == None
```
mkrets, line 130: There seems to be an extra space in each line.

mkrets:
I can confirm that this PR captures the relevant changes.
Thanks for looking into this!

PR #584, opened 2020-10-14
Adaptahop support longint

cphyc:
Add support for reading Adaptahop catalog generated with the `LONGINT` flag (i.e. those for which the particle ids may be larger than `2**31-1 = 2147483647`.

### DIFF
--- a/pynbody/halo/adaptahop.py
+++ b/pynbody/halo/adaptahop.py
@@ -14,31 +14,32 @@ from ..extern.cython_fortran_utils import FortranFile
 
 
 class DummyHalo(object):
-
     def __init__(self):
         self.properties = {}
+
 
 unit_length = units.Mpc
 unit_vel = units.km / units.s
 unit_mass = 1e11 * units.Msol
 unit_angular_momentum = unit_mass * unit_vel * unit_length
-unit_energy = unit_mass * unit_vel**2
+unit_energy = unit_mass * unit_vel ** 2
 unit_temperature = units.K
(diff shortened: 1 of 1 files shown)


pynbody/halo/adaptahop.py
```diff
@@ -136,10 +142,35 @@ def _read_halo_data(self, halo_id, offset):
...
+        possible_dtypes = list({"i", "q"} - {default_dtype})
+
+        dtypes = [default_dtype] + possible_dtypes
+
+        def _helper(fpu, expected_size):
+            ipos = fpu.tell()
+
+            for dtype in dtypes:
+                try:
+                    iord_array = fpu.read_vector(dtype)
```
apontzen, line 156: Is this a bit strange - repeatedly reading the file and allowing it to fail? Would it be easy/more streamlined to actually look at the fortran reported block size and infer the dtype from that directly?

cphyc: This code should only fail on the first read. After this, the right dtype is stored (so there should not be any significant performance hit, and worst case is that the data will be read twice).

It is possible in principle to read the header of the Fortran record and infer the dtype given the expected number of elements to be read, but it requires having a fairly low-level piece of code there. Another option is to add support for this directly in the Fortran file reader (in `fpu.read_vector`) by adding the possibility to specify a length and possible data types.

apontzen: OK, fair enough! Thanks

PR #587, opened 2020-11-17
Massform read

emapple:
Many tipsy simulations include massform values in both the starlog and an auxiliary array. Pynbody will by default read the massform values from the starlog, though reading them from the auxiliary array is faster (especially in the case of single halos).

I have changed the default to reading from the auxiliary array, if it exists.


apontzen:
Thanks!

PR #589, opened 2020-11-20
Fix issue 588

### LINKED ISSUE #588: tipsy starlog metadata read error when starlog data listed multiple times in log file
When restarting changa from an output, the full parameter list plus starlog metadata is written another time. The pynbody starlog reader should stop once it hits the end of the first metadata list. As it is, the multiple reads results in an incorrect file format and a value error is raised saying that the metadata read has failed.

emapple:
Fix #588 (make sure to only read one set of metadata if duplicates exist)

Also, strips leading whitespace to allow for misprints in the log file


apontzen:
Thanks!

PR #591, opened 2020-12-17
Fix integration limit in projected kernel

elehcim:
Found this bug while using pynbody to create SPH images.

The integration limit of the projected kernel should be `2*h`.
For a proof relying on integrating analytically the spline kernel see here (with help of Sven De Rijcke and @shivangeerathi): https://github.com/elehcim/projected_kernel/blob/main/projectedkernel.ipynb

Probably some tests should be fixed too.


apontzen:
Thanks!

Issue #581, opened 2020-09-07
Writing profile to disk fails because hashlib does not support usedforsecurity keyword


Martin-Rey:
Hi all,

I am trying to write a `pynbody.profile.Profile` object, using its build-in method `write()`:
https://github.com/pynbody/pynbody/blob/6b0ded6b3f21d83b86143c391298bfcb3934b245/pynbody/analysis/profile.py#L481
 which fails with the following trace:
```
Traceback (most recent call last):
  File "/Users/martinrey/Documents/stellar_halos/plot_surface_brightness.py", line 70, in <module>
    profile2d.write()
  File "/opt/anaconda3/lib/python3.7/site-packages/pynbody-0.48-py3.7-macosx-10.9-x86_64.egg/pynbody/analysis/profile.py", line 551, in write
    filename = self._generate_hash_filename()
  File "/opt/anaconda3/lib/python3.7/site-packages/pynbody-0.48-py3.7-macosx-10.9-x86_64.egg/pynbody/analysis/profile.py", line 522, in _generate_hash_filename
    (hashlib.md5(self._x, usedforsecurity=False)).hexdigest()
TypeError: openssl_md5() takes no keyword arguments
```

Thanks to https://stackoverflow.com/questions/54717862/how-do-i-know-if-the-usedforsecurity-flag-is-supported-by-hashlib-md5 and https://github.com/sphinx-doc/sphinx/issues/7611, I have tracked this to the fact that my platform (Mac OS) does not support the `usedforsecurity` keyword used in the hashing here:
https://github.com/pynbody/pynbody/blob/6b0ded6b3f21d83b86143c391298bfcb3934b245/pynbody/analysis/profile.py#L468-L477

From python, `hash lib` is indeed missing the keyword
```
In [1]: import hashlib, inspect                                                                                                        

In [2]: inspect.getfullargspec(hashlib.new)                                                                                            
Out[2]: FullArgSpec(args=['name', 'data'], varargs=None, varkw='kwargs', defaults=(b'',), kwonlyargs=[], kwonlydefaults=None, annotations={})
```

for the following platform configuration
```
In [3]: import sys, platform, pynbody, numpy; print("%s\r\n%s\r\npynbody %s\r\nnumpy %s\r\n"%(sys.version, platform.platform(), pynbody
   ...: .__version__, numpy.__version__))                                                                                              
3.7.4 (default, Aug 13 2019, 15:17:50) 
[...]

PR #599, opened 2021-02-11
Fix hashing and profile writing

Martin-Rey:
I believe this should fix issue #581 and adds tests to verify the profile writing ability.

I had a number of python 2-3 compatibility issues when fixing this code (binary file reading convention, string encoding in utf-8), which I am not too confortable about, feel free to comment on anything.

I checked that the unique hash generation is conserved after these changes, so the code should stay backward compatible for people with already generated profiles on disk.

Martin


apontzen:
Looks great to me. Thanks Martin!

Issue #530, opened 2019-06-15
Bug in __len__ method of GrpCatalogue 


Martin-Rey:
Hi Andrew,

This is a follow-up to the issue first raised here (https://github.com/pynbody/tangos/issues/108). There is an inconsistency in the index structure of HOP halo catalogue. Their recorded length as evaluated by __len__ is one less than what their index structure suggests.

I tracked it to the override of the __len__ method of `GrpCatalogue`, from which HOP inherits:
https://github.com/pynbody/pynbody/blob/b37d0dba99d7b64ab2fc51355d8068afb609d7ab/pynbody/halo/__init__.py#L188-L197

`self.base[self._array]` produces an array with the length of the parent simulation and containing the number of the group to which each particle belongs. `max()` will then rightfully return the total number of groups if group numbers run from 1 to N.

In the case of HOP for which indeces run from 0 to N-1, the method will wrongfully return a length of N-1 and miss the last halo from the catalogue.

Hope this is clear,
Martin

PR #600, opened 2021-02-12
Ensuring HOP catalogues have correct lengths

Martin-Rey:
Ensures that the HOP halo catalogue returns a consistent length compared to its index structure (issue #530). This ensures that cycling from 0 to __len__ does not miss the last halo in the catalogue (https://github.com/pynbody/tangos/issues/108) 

I verified the behaviour is correct manually, but I cannot find a HOP catalogue to test on in the testdata. Is there one I am missing somewhere?

Martin

### DIFF
--- a/pynbody/halo/hop.py
+++ b/pynbody/halo/hop.py
@@ -23,7 +23,7 @@ class HOPCatalogue(GrpCatalogue):
                 raise RuntimeError("Unable to find HOP .tag file in simulation directory")
 
         sim._create_array('hop_grp', dtype=np.int32)
-        sim['hop_grp'] -= 1
+        sim['hop_grp'] = -1
         with open(fname, "rb") as f:
             num_part, = struct.unpack('i', f.read(4))
             if num_part == 8:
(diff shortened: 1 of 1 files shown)


pynbody/halo/hop.py
```diff
@@ -21,22 +22,28 @@ def __init__(self, sim, fname=None):
             if not os.path.exists(fname):
                 raise RuntimeError("Unable to find HOP .tag file in simulation directory")
 
-
         sim._create_array('hop_grp', dtype=np.int32)
-        sim['hop_grp']=-1
+        sim['hop_grp'] -= 1
```
apontzen, line 26: This version appears to assume that `hop_grp` is initialised to all zeros. That sounds like an unsafe assumption? Is there a reason why setting it to `-1` (the original behaviour) is a problem?

apontzen:
I don't think there is one... I suggest treating the lack of a test as a separate issue and merging this for now (modulo my query above)

pynbody/halo/hop.py
Martin-Rey: Yes sorry this is a typo. This should go back to the original -1.

Martin-Rey:
Agreed, also generating a HOP test needs a full snapshot (it would not work on the partial test outputs). Given how small this correction is, I'm happy to merge as is and leave the implementation of a test to more significant design changes.

PR #601, opened 2021-02-15
Fix neighbour list python interface

elehcim:
I refreshed the methods `kdtree.nn()` and `kdtree.all_nn()`.

Looking at the git blame, I think the code hasn't been updated after kdtree got multithreading support.
I had to dig into the (undocumented) cpp code to try to fix the functionality. The main changes are:

* The signature of `kdmain.nn_start` was not coherent.
* the value returned by `smSmoothStep` wasn't checked properly in `kdmain.nn_next`.
* An initialization `kdmain.domain_decomposition` step was missing.

The last two points probably are an heritage of when multithreading was not implemented.

This PR may help to answer these [two](https://groups.google.com/g/pynbody-users/c/rVuPKYwa1xY/m/suthSAyjAQAJ) [questions](https://groups.google.com/g/pynbody-users/c/yHMYYznGF0k).

### DIFF
--- a/nose/sph_smooth_test.py
+++ b/nose/sph_smooth_test.py
@@ -114,8 +114,36 @@ def test_periodic_smoothing():
     npt.assert_allclose(f.dm['smooth'][::100],
                          np.load('test_smooth_periodic.npy'),rtol=1e-5)
 
-    npt.assert_allclose(f.dm['rho'][::100],
-                         np.load('test_rho_periodic.npy'),rtol=1e-5)
+def test_neighbour_list():
+    f = pynbody.load("testdata/test_g2_snap")
+    pynbody.sph._get_smooth_array_ensuring_compatibility(f.g)  # actual smoothing
+    t = f.g.kdtree
+    n_neigh = 32
+
+    generator_nn = t.nn(n_neigh)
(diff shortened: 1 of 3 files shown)


elehcim:
Looking at the test I saw the distance of particle 9 to neighbour list particle 9 is not zero. Because the labeling is not coherent. I think it makes sense to improve the neighbour list labeling, returning the particle iOrder there too... Needs fixing.

elehcim:
Done, let's wait for travis

apontzen:
Thanks! This was never intended to be a public interface, but now that you have made it work, could you add a couple of lines of documentation?

The parallelisation of this was always pretty naive, but I never pursued it much because for the main use case (smoothing) it did not dominate the runtime. However for this use scenario, it may be more problematic. Did you look into it at all?

elehcim:
> The parallelisation of this was always pretty naive, but I never pursued it much because for the main use case (smoothing) it did not dominate the runtime. However for this use scenario, it may be more problematic. Did you look into it at all?

No, I didn't. For my use cases the neighbours retrieval is quite fast though. I understand that probably in the future a look into that would be required.

apontzen:
Thanks, great stuff!

Issue #598, opened 2021-02-11
Inconsistent temperature, molecular weight and sound speed derivations for Ramses snapshots


Martin-Rey:
Hi,

I noticed an inconsistency in the way temperature, molecular weight and sound speeds are derived for Ramses snapshots.

For traditional snapshots, pynbody derives a molecular weight field `sim.g['mu']` , assuming a steep transition from 0.6 for gas above 10^4 K, to 1.3 for colder gas https://github.com/pynbody/pynbody/blob/d9124e7b76d6f768e707d1aee0c45d2303ce9162/pynbody/derived.py#L229-L244

While this is approximate, it at least captures the difference between ionized and neutral gas. This difference is then mapped self-consistently when deriving the temperature `sim.g['temp']` from the internal, stored hydro quantities (energy, rho or pressure)
https://github.com/pynbody/pynbody/blob/d9124e7b76d6f768e707d1aee0c45d2303ce9162/pynbody/derived.py#L262-L271
and then in the sound speed:
https://github.com/pynbody/pynbody/blob/d9124e7b76d6f768e707d1aee0c45d2303ce9162/pynbody/derived.py#L222-L225

For Ramses simulations, the molecular weight is derived the same way, but the `sim.g['temp']`, however, is redefined at the bottom of the snapshot file, assuming a constant and never evolving molecular weight of 1.22 (the neutral gas one):
https://github.com/pynbody/pynbody/blob/d9124e7b76d6f768e707d1aee0c45d2303ce9162/pynbody/snapshot/ramses.py#L1144-L1146
This leads to a factor of 2 offset in the temperature for ionised gas, which is overestimated by 1.22 / 0.6. This offset then propagates to the sound speed calculation.

I will submit a fix that at least uses the self-consistent, pynbody derived, molecular weight field for the temperature. A more long-term and precise solution would be to read the cooling tables outputted by Ramses, and derive the precise molecular weight at each temperature and gas density as computed internally by the code. 

Cheers,
Martin

Martin-Rey:
Ramse is now consistent with pynbody through #602, although a more generally problematic behaviour across all snapshot types is described in #605

Issue #597, opened 2021-02-08
Detecting Ramses cosmological runs without namelist.txt


Martin-Rey:
Hi @mkrets  and @apontzen,
 
Since #534 and #582, the detection of cosmological runs with Ramses relies exclusively on opening and reading the namelist:
https://github.com/pynbody/pynbody/blob/d9124e7b76d6f768e707d1aee0c45d2303ce9162/pynbody/snapshot/ramses.py#L996-L1000
when it previously relied on cosmological parameter information. While reading the namelist is a technical improvement, it happens quite often that the namelist.txt, patches.txt or compilation.txt files from Ramses get corrupted on supercomputers, and are unreadable. In such a case, pynbody will throw the following user warning
```
/cosma6/data/dp171/shared/python/anaconda3/lib/python3.8/site-packages/pynbody-1.0.2-py3.8-linux-x86_64.egg/pynbody/snapshot/ramses.py:537: UserWarning: Namelist found but unable to read.
  warnings.warn("Namelist found but unable to read.")
```
and conclude that the run is not cosmological. This leads to negative simulation ages, even if cosmogical parameters are read in and one can compute the correct `age' with analysis routines
```
In [1]: import pynbody                                                                                                                                                                                                    

In [2]: s = pynbody.load("../Halo383_fiducial/output_00060")                                                                                                                                                              
/cosma6/data/dp171/shared/python/anaconda3/lib/python3.8/site-packages/pynbody-1.0.2-py3.8-linux-x86_64.egg/pynbody/snapshot/ramses.py:537: UserWarning: Namelist found but unable to read.
  warnings.warn("Namelist found but unable to read.")
/cosma6/data/dp171/shared/python/anaconda3/lib/python3.8/site-packages/pynbody-1.0.2-py3.8-linux-x86_64.egg/pynbody/snapshot/ramses.py:793: UserWarning: Using field 'tform' to distinguish stars. If this is wrong, try editing your config.ini, section [ramses], entry particle-distinguisher.
  warnings.warn("Using field %s to distinguish stars. If this is wrong, try editing your config.ini, section [ramses], entry particle-distinguisher."%pb_name)

In [3]: s.physical_units()                                                                                                                                                                                                

In [4]: s.properties['time']                                                                                                                                                                                              
Out[4]: Unit("-4.09e+00 kpc s km**-1")

In [5]: s.properties['time'].in_units("Gyr")                                                                                                                                                                              
Out[5]: -3.998882056560654

[...]

Issue #554, opened 2020-02-17
Broken Ramses birth times for isolated simulations


Martin-Rey:
Hi,

A master student in Lund, Oscar Andersson, made me aware that the current handling of 'ages' and 'tform' arrays for RAMSES isolated, non-cosmological, simulations is broken, most likely due to PR #521 . 

By default, pynbody now tries to load the birth files to convert the birth times of stellar particles in a physically-meaningful unit, but this operation (and the associated birth files) are unnecessary in the context of isolated simulations.

I will submit a fix refining the behaviour here:
https://github.com/pynbody/pynbody/blob/e0ade3bf17eaa2d9d3eb2aa63853ed6c48452c6e/pynbody/snapshot/ramses.py#L971-L977

Is there an isolated snapshot in old-RAMSES format in the current test data to design a test?

Martin

apontzen:
Thanks Martin. I don't think there is such a snapshot in the test data but I'd be happy to add one if you have access to a small one.

Martin-Rey:
Should be fixed by #602

Issue #542, opened 2019-10-16
Ramses-RT photon density unit


Martin-Rey:
Hi all, 

I am not entirely sure whether this is bug or something I do not yet understand. 

When using pynbody on a RAMSES-RT run and loading `rad_0_rho` or `rad_0_flux_x` (e.g. for the photon group 0 but this is true for my 6 photon groups), both quantities come out with the same unit equivalent to flux (length^-2 * time^-1):

```
In [1]: srt = pynbody.load("./HaloXXX_RT/output_00080")

In [2]: srt.physical_units()

In [3]: srt.g['rad_4_rho']
Out[3]: 
SimArray([3.15841890e+54, 3.14141689e+54, 3.14210700e+54, ...,
          2.04582717e+54, 2.09797356e+54, 2.40695327e+54], 'km s**-1 kpc**-3')

In [4]: srt.g['rad_4_flux_x']
Out[4]: 
SimArray([-8.04207711e+51,  7.73927157e+53,  4.03112470e+53, ...,
          -8.31079212e+52, -8.61794973e+52, -1.08170997e+53], 'km s**-1 kpc**-3')
```

I am slightly perplexed by these units, as I would have expected photon densities (rad_X_rho) to be in units of length*-3. 

RAMSES-RT outputs the photon densities "in units of flux", but shouldn't PYNBODY re-instore the missing factor of `c=speed of light` when loading the unit to make it physical? Or is there a reason to want photon densities in flux units?

Best,
Martin

Martin-Rey:
Following a discussion with @jokirosdahl, the photon density is stored on disk with flux units to take out the dependency on the "reduced speed of light" of each run.

Re-instating the physical number in photons/cc cube would require to infer the value of the reduced speed of light from the namelist (rt_c_cgs or rt_c_fraction parameters). 

I think this is a task that is best left to the user, given the potential for introducing bugs in the guessing process or if the namelist cannot be found or read. Let me know if you agree with this assessment, and whether we should leave this issue open.

apontzen:
Thanks! I guess it would be neat to solve this but agree it could introduce bugs. Unless someone's prepared to do some work, I'd suggest maybe leaving a comment in the code somewhere obvious?

PR #602, opened 2021-02-16
Ramses loader fixes

Martin-Rey:
Hi,

I believe this PR several problems with the Ramses loader that were left standing for a while:

#542: missing physical unit for photon densities in RT runs --> the loader now explicitly warns the user about this missing factor and is tested for warning generation

#554: broken tform for isolated simulations --> loader now tests whether run is cosmoligical before attempting to read the birth files and transforming the `tform` array. Isolated simulations are correctly left untouched and both use cases are tested.

#597: detecting cosmological runs --> loader reinstates the old mechanism to detect cosmological runs, which was more robust than the namelist detection.  It generating a warning when namelist cannot be read or is not found for inferring a cosmological run.

#598: inconsistent `mu` in temperature calculation --> loader now uses the pynbody `mu` field rather than the fixed 1.22. The redefinition of the temperature in the Ramses loader is in fact needed as the traditional definitions generate Circular dependencies. This fix is to wait while the implementation reading the cooling table to infer the molecular weight is ready

Any comments welcome
Martin

### DIFF
--- a/pynbody/snapshot/ramses.py
+++ b/pynbody/snapshot/ramses.py
@@ -994,10 +994,17 @@ class RamsesSnap(SimSnap):
         return list(keys_ND)
 
     def _not_cosmological(self):
+        not_cosmological = True
+
         if "cosmo" in self._namelist and self._namelist["cosmo"]:
-            return False
-        else:
-            return True
+            not_cosmological = False
+
+        if not self._namelist:
+            warnings.warn("Namelist file either not found or unable to read" +
+                          "guessing whether run is cosmological from cosmological parameters")
+            not_cosmological = (self._info['omega_k'] == self._info['omega_l'] == 0)
+
+        return not_cosmological
 
     def _convert_tform(self):
         # Copy the existing array in weird Ramses format into a hidden raw array
@@ -1006,7 +1013,7 @@ class RamsesSnap(SimSnap):
 
         if self._is_using_proper_time:
             t0 = analysis.cosmology.age(self, z=0.0, unit="Gyr")
-            birth_time = t0 + self.s["tform_raw"].in_units("Gyr")/self.properties["a"]**2
+            birth_time = t0 + self.s["tform_raw"].in_units("Gyr") / self.properties["a"] ** 2
             birth_time[birth_time > t0] = t0 - 1e-7
             self.star['tform'] = birth_time
         else:
--- a/pynbody/snapshot/ramses.py
+++ b/pynbody/snapshot/ramses.py
@@ -1000,30 +1000,28 @@ class RamsesSnap(SimSnap):
             not_cosmological = False
 
         if not self._namelist:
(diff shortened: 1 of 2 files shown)


pynbody/snapshot/ramses.py
```diff
@@ -1143,4 +1155,13 @@ def mass(sim):
...
 @RamsesSnap.derived_quantity
 def temp(sim):
-    return ((sim['p'] / sim['rho']) * (1.22 * units.m_p / units.k)).in_units("K")
+    """ Gas temperature derived from pressure and density """
+    # Has to be redefined and rederived here from Ramses native variables
+    # to avoid running into circular dependencies with the traditional derived definition
+    # Now uses the self-consistent molecular weight field from pynbody (issue 598)
+    from ..derived import mu
+    mu_est = array.SimArray(np.ones(len(sim)), units="1")
+    for i in range(5):
```
apontzen, line 1164: Why 5?

Martin-Rey: This is matching the existing derived quantity:
https://github.com/pynbody/pynbody/blob/b7bb374404521f3b9922df52dd3d4bd6b2e5ab7e/pynbody/derived.py#L268-L277

I am not entirely sure why 5 iterations were deemed enough or not for the base property.

apontzen: I see... could you fix it in both places? I think there should be a convergence check to a defined tolerance (maybe +/- 0.1%?) and also an upper limit on the number of iterations, with a warning issued if the solution does not converge in time.

Martin-Rey: Sure that sounds good, I can include this this in. Is it ok to do it within this PR or would you want it as a separate fix?

apontzen: Thanks! Fine just to do it here I think.

Martin-Rey: I am having issues with implementing a convergence to a given tolerance. Due to the binary nature of the default mu implementation (either 1.3 or 0.59 depending on a temperature cut to 1e4K):
https://github.com/pynbody/pynbody/blob/b7bb374404521f3b9922df52dd3d4bd6b2e5ab7e/pynbody/derived.py#L235-L250
one cannot actually converge for values around between 5e3 K and 2e4K as the molecular weight just oscillates between 1.3 and 0.59 which each iteration. 

The only solution I see is to improve the actual implementation of `mu` to avoid the binary cut and have a function of temperature that can actually converge. I suggest this is re-opened as an independent issue, as this is global to pynbody rather than just Ramses? 

Any insights?

apontzen: Yes I agree this sounds like it deserves its own issue

Martin-Rey: Done in #605, any other comments on this?

PR #606, opened 2021-03-25
Gadget 4 and arepo HDF files support

apontzen:
Gadget 4 and arepo change lots of details about HDF files. This is an attempt to support them while
maintaining back-compatibility with older versions of the Gadget HDF snapshot.


regmachado:
It seems that in Gadget-4 the Omegas are in the Parameters, bu the 'Time' attribute is still in the Header. So when 'Time' is not found in the Parameters, age from cosmology is apparently being calculated even in non-cosmological snapshots.

bwkeller:
I think this PR supercedes my #569, apologies for forgetting about that.

apontzen:
Oh whoops! Reinventing the wheel. When you say it supercedes #569, is there anything this does that #569 does not?

bwkeller:
Let me take a look, I think this covers everything but I should actually
compare to make sure the AREPO outputs I have behave the same in both.

bwkeller:
Finally got around to testing this out, and it does do everything #569 does.  I've closed #569, so if you want to merge this one now feel free.  Thanks @apontzen and everyone!

PR #607, opened 2021-03-26
More flexibility when reading particle files with RAMSES

cphyc:
When using MC tracer, the local random seed is dumped together with the seed used for tracers. This causes the record 'randseed' to have a length of 8bytes instead of 4.
This fixes it by skipping the record altogether.


cphyc:
This is related to the recent merge of the MC tracer particle in RAMSES https://bitbucket.org/rteyssie/ramses/pull-requests/456.

PR #614, opened 2021-05-20
Read HDF outputs of Subfind in Gadget-4

lluciesmith:
Added support to read HDF outputs of Subfind in Gadget4. I have also included tests of the pipeline in the nosetest directory. This implementation currently only supports reading single-file outputs (and not multi-file) produced by Subfind.


nose/utils_test.py
```diff
@@ -123,10 +123,12 @@ def test_openmp_summations():
     start = time.time()
     sum_a = np.dot(a,(b>1.0))
     np_timing = time.time()-start
+    print(sum_a)
```
apontzen, line 126: Could you revert these two print statements unless there is a reason for them to be present?

nose/subfindhdf_gadget4_test.py
```diff
@@ -0,0 +1,317 @@
...
+        assert(np.allclose(halos[hid].dm['iord'], htest[hid]['iord']))
+
+
+import six
+from os.path import isfile
+import numpy as np
+import h5py
+
+
+class Halos:
```
apontzen, line 75: Would be useful to add a docstring here to explain what it is doing in the context of this test

pynbody/halo/subfindhdf.py
```diff
@@ -310,4 +314,186 @@ def __getattr__(self, name):
...
+        return header
+
+    def _read_ids(self):
+        data_ids = np.array([], dtype=self.dtype_int)
+        for n in range(self._tasks):
+            ids = self.base._hdf_files[n][self.ptype]['ParticleIDs']
+            data_ids = np.append(data_ids, ids)
+        return data_ids
+
+    def _read_groups(self):
```
apontzen, line 432: Could you add a couple of comments or split up into a couple of methods to clarify what is happening here? (I realise other loaders are not great at this but we might as well improve the code standard.)

apontzen (commented):
Thanks Luisa! I have a couple of small comments here. Happy to merge after that assuming the tests pass (I just restarted them with the new test data tar).

nose/utils_test.py
lluciesmith: Yes I can revert them. The reason why they are present is that on my local machine that test fails without the print statements (I never really understood why..). So I always have to manually add those print statements to run the nosetests. But since this problem is unrelated to this specific pull request, I can delete the print statements.

lluciesmith:
The tests seem to have failed because it can't find the snapshot. Does the path that I wrote in the test match the location of the snapshots?

apontzen:
Thanks!

PR #626, opened 2021-10-18
Also test on Python 3.10

cphyc:
This activates CI tests on Python 3.10 (upcoming stable version).

Keeping this in "work in progress" because at the moment all dependencies have to be compiled (numpy, matplotlib, ...), and that makes the tests extremely slow. If you don't think this is a problem, I'd be happy to merge it as is.


cphyc:
Closing as this won't fix until we migrate to pytest.

PR #630, opened 2021-11-29
AdaptaHOP: use SimArray instead of Units

cphyc:
This PR makes it so that the AdaptaHOP frontend returns `SimArray` instances instead of `Unit` instances. It also replaces non-explicit names with more explicit ones. This is in support of https://github.com/pynbody/tangos/pull/160.


pynbody/halo/__init__.py
```diff
@@ -56,6 +57,24 @@ def is_subhalo(self, otherhalo):
 
         return self._halo_catalogue.is_subhalo(self._halo_id, otherhalo._halo_id)
 
+    def physical_units(self, distance='kpc', velocity='km s^-1', mass='Msol', persistent=True):
```
apontzen, line 60: There is something a little bit weird about this. If I understand correctly, you are intervening when somebody calls `physical_units` on a halo, to ensure that its properties are also put into physical units. However, I am a bit nervous:

* If `h` is a halo catalogue and I ask for `h[1].physical_units()`, then the arrays of  all halos are converted into physical units (this is unavoidable) but only the properties of halo 1?
* If `f` is the underlying snapshot, `f.physical_units()` doesn't convert any of the halo properties, and indeed there is seemingly no way to convert all of the halo properties

I wonder whether a more appropriate thing to do would be for the halo catalogue `h` to have a `physical_units` method which

* by default, calls `f.physical_units()`
* where supported, converts all the halo properties then calls `f.physical_units()`
* can be called by `h[i].physical_units()`

As a final thought, I see a copy/paste of the code to determine `new_unit` which should instead be factored out somewhere.

cphyc: I agree with all your points. However, there is a technical difficulty here which is that the halo properties are computed just-in-time, so I should need to store somewhere the fact that `f.physical_units()` has been stored. Do you have an idea of how to implement this in a clean way?

apontzen: Yes, there is already a mechanism for this. The snapshot has a variable `_autoconvert` which either contains `None` (for no conversion) or the set of units if `physical_units` has been called previously (with `persist=True`). I think the halos should be able to examine that variable at the point of lazy-computation of the halo properties?

cphyc: This should be good now as per 0df689f and 4fa96b1

pynbody/halo/__init__.py
```diff
@@ -56,12 +62,27 @@ def is_subhalo(self, otherhalo):
 
         return self._halo_catalogue.is_subhalo(self._halo_id, otherhalo._halo_id)
 
+    def physical_units(self, distance='kpc', velocity='km s^-1', mass='Msol', persistent=True, convert_parent=True):
+        if convert_parent:
```
apontzen, line 66: To make sure I understand this distinction:

* with `convert_parent=True`, all arrays in the parent convert, and each subhalo (including this one) will have `physical_units` called with `convert_parent=False`
* with `convert_parent=False`, only this one halo has its properties converted and arrays are ignored

I find the comment here a bit hard to parse — I guess you are drawing attention to why `_autoconvert_properties` is _not_ called in the first code path? Maybe just find a way to make this comment a bit clearer.

pynbody/snapshot/__init__.py
```diff
@@ -35,8 +35,105 @@
 
 logger = logging.getLogger('pynbody.snapshot')
 
+class UnitFullContainer:
```
apontzen, line 38: I like this solution but could it be in its own file? The `snapshot/__init__.py` is already far too long and diverse.

Also I wonder whether the name is clear? Perhaps something like `ContainerWithPhysicalUnitsOption`, while clunky, is clearer?

pynbody/snapshot/__init__.py
```diff
@@ -35,8 +35,105 @@
...
+            return
+
+        all = list(self._arrays.values())
+        for x in self._family_arrays:
+            all += list(self._family_arrays[x].values())
+
+        for ar in all:
+            self._autoconvert_array_unit(ar.ancestor, dims)
+
+    def physical_units(self, distance='kpc', velocity='km s^-1', mass='Msol', persistent=True):
```
apontzen, line 106: Should you add and document the `convert_parent` kwarg here, to save potential confusion? It can simply have no effect if there is no parent object.

pynbody/snapshot/__init__.py
```diff
@@ -35,8 +35,105 @@
...
+
+        **Optional Keywords**:
+
+           *distance*: string (default = 'kpc')
+
+           *velocity*: string (default = 'km s^-1')
+
+           *mass*: string (default = 'Msol')
+
+           *persistent*: boolean (default = True); apply units change to future lazy-loaded arrays if True
```
apontzen, line 121: Since I understand that properties, as well as arrays, may now be auto-converted should this line be clarified?

apontzen (commented):
Some final thoughts — and also there is now a conflict with master due to the switch to pytest. Hopefully all this will be easy for you to address.

Ping me when you’re ready and I will merge right away.

pynbody/snapshot/__init__.py
cphyc: My understanding is that properties are static and not lazy-loaded, so they are all converted immediately and the persistency is a consequence of that, isn't it?

apontzen: OK! I misunderstood one of your previous comments to mean that some halo properties in adaptahop catalogues end up being lazy loaded.

pynbody/halo/__init__.py
cphyc: I have added a note in the docstring above. Is now clearer?

pynbody/snapshot/__init__.py
cphyc: I added docstrings to `physical_units` with a different explanation for `persistent` depending on context. Tell me what you think

pynbody/snapshot/__init__.py
cphyc: Done

pynbody/halo/__init__.py
apontzen: Yes I think that's about as clear as it can be.

However there is a typo in your documented default value for `convert_parent` (it says `None` but the actual default is `True`)

PR #631, opened 2021-11-29
Migrate to pytest

cphyc:
nose is not being developed anymore. In particular, it doesn't work with Python 3.10 (see #626) and will probably never work with future versions.

This is a first stab at migrating to pytest.

Note that pytest reports all warnings it encounters and has the ability to treat them as errors. I haven't activated this functionality at the moment, but it allowed me to catch many of them (in part due to some numpy deprecated features).
I have also replaced `type(foo) is Bar` by `isinstance(foo, Bar)` where it seemed correct to do so.


.github/workflows/build-test.yaml
```diff
@@ -24,32 +24,49 @@ jobs:
...
+      run: python -m pytest tests/ramses_new_ptcl_format_test.py tests/ramses_test.py --junitxml=pytest_noshmem.xml
 
+    - name: Upload Unit Test Results
+      if: always()
+      uses: actions/upload-artifact@v2
+      with:
+        name: Unit Test Results (Python ${{ matrix.python-version }})
+        path: ${{ github.workspace }}/**/pytest*.xml
+
+  event_file:
```
cphyc, line 64: As a bonus: it is relatively easy to upload the build results of pytest to Github actions using the following (and the `upload-unit-test-results.yaml` workflow below).

pynbody/bridge/_bridge.pyx
```diff
@@ -50,8 +50,7 @@ def bridge(npc.ndarray[integral_1, ndim=1] iord_to,
             output_index[i] = 0
             found_match[i] = 0 # false
 
-    return (output_index, found_match.astype(np.bool))
-
+    return (output_index, found_match.astype(bool))
```
cphyc, line 53: Note: `np.bool` and `np.int` have been deprecated in favour of `bool` and `int` respectively.
See https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated.

pynbody/gravity/__init__.py
```diff
@@ -1,5 +1 @@
-import imp
-import numpy as np
-
 from . import calc
-imp.reload(calc)
```
cphyc, line 5: The `imp` library has been deprecated in favour of `importlib`, but I couldn't find any reason why we were reloading the .so files?

pynbody/halo/ahf.py
```diff
@@ -363,7 +363,7 @@ def _load_ahf_halos(self, filename):
         f = util.open_(filename,"rt")
         # get all the property names from the first, commented line
         # remove (#)
-        keys = [re.sub('\([0-9]*\)', '', field)
+        keys = [re.sub(r'\([0-9]*\)', '', field)
```
cphyc, line 366: Here and in the following, there were many warnings due to invalid escape sequences (e.g. `\(` is an invalid escape sequence). This is fixed by adding the `r` before the string.

pynbody/halo/rockstar.py
```diff
@@ -243,7 +243,7 @@ def __init__(self, sim, dummy=False, filename=None, format_revision=None, **kwar
             raise IOError("Halo catalogue not found -- check the file name of catalogue data or try specifying a catalogue using the filename keyword")
 
         with f:
-            self._head = np.fromstring(f.read(self.head_type.itemsize),
+            self._head = np.frombuffer(f.read(self.head_type.itemsize),
```
cphyc, line 246: The numpy doc recommends using `np.frombuffer` when reading-in binary data (as is the case here). See https://numpy.org/doc/stable/reference/generated/numpy.fromstring.html.

pynbody/halo/ahf.py
```diff
@@ -317,7 +317,7 @@ def _load_ahf_particle_block(self, f, nparts=None):
             if self._use_iord:
                 data = self._iord_to_fpos[data]
             else:
-                if type(self.base) is not snapshot.nchilada.NchiladaSnap:
+                if not isinstance(self.base, snapshot.nchilada.NchiladaSnap):
```
cphyc, line 320: There are multiple places where the syntax `if type(foo) is Bar` was used, where it could be `if isinstance(foo, Bar)`.
The former does not work with subclasses while the latter do. I have replaced the ones that appeared to fall in the second category.

pynbody/snapshot/__init__.py
```diff
@@ -957,23 +957,23 @@ def _transform(self, matrix):
     def rotate_x(self, angle):
         """Rotates the snapshot about the current x-axis by 'angle' degrees."""
         angle *= np.pi / 180
-        return self.transform(np.matrix([[1,      0,             0],
-                                         [0, np.cos(angle), -np.sin(angle)],
-                                         [0, np.sin(angle),  np.cos(angle)]]))
+        return self.transform(np.array([[1,      0,             0],
```
cphyc, line 960: From numpy's doc (https://numpy.org/doc/stable/reference/generated/numpy.matrix.html):
>It is no longer recommended to use this class, even for linear algebra. Instead use regular arrays. The class may be removed in the future.

tests/snapshot_test.py
```diff
@@ -103,12 +95,8 @@ def test_derived_array():
     f['vr']
     assert 'vr' in f.keys()
 
-    try:
+    with pytest.raises(ValueError):
```
cphyc, line 98: The following line only raises a `ValueError`, and neither a `RuntimeError` or a `TypeError`, so only catch the former.

tests/sph_image_test.py
```diff
@@ -21,7 +26,23 @@ def setup():
     f.gas['smooth']=(f.gas['mass']/f.gas['rho'])**(1,3)
     f.physical_units()
 
-def test_images():
+@pytest.fixture
```
cphyc, line 29: Note: in this file, the tests use fixtures (which are passed as arguments to the test functions). Note that I could also have updated the old versions to use the absolute path (`test_folder / "..."`) with the same end result.

apontzen:
Thanks for all this work. I don't propose to do a complete line-level review since it is such a big diff. Are there particular things you think would be worth checking over?

cphyc:
There are two thinkgs:
1. A "breaking" change is the replacement of the `type(something) is SomeClass` into `isinstance(something, SomeClass)`. There were locations where the latter seemed to be preferable (as it allows for subclasses to pass the test).
2. We should make sure that the PR hasn't removed any test (i.e. the number of tests now equal the number of tests before). This should be fine as far as my verifications went, but it may be useful to check it independently?

cphyc:
@apontzen I rebased this PR to remove as much as possible and get a more minimalistic migration. This should now be reasonably easier to go through, but note that the warnings I was fixing previously aren't fixed anymore (it'll come in a second batch).

apontzen:
@cphyc Thanks! I had a look through the changes and verified that (1) the number of tests hasn't changed and (2) I agree with your reasoning on the potentially breaking changes you mentioned. So I'll be happy to merge this once you are ready.

cphyc:
It's ready to go on my side!

PR #633, opened 2021-12-01
Lazy RAMSES & AdaptaHOP loading

cphyc:
This allows loading faster RAMSES snapshot by:

- delaying the particle counting to when it is actually required (i.e. quite late);
- use custom-made routines to count the number of gas cells and particles that do this and only this (instead of relying of general-purpose iterators that also read particles families or cell positions).

The above improvement allows to load AdaptaHOP catalogs much faster
- implemented a `index_parent` kwa on creation of an AdatapHOP dataset that allows loading quickly the halo catalogue (but halo objects don't have access to their parent).

| Action                        | Timing (`master`)  | Timing (`f.halo(index_parent=False)`) | Timing (`f.halo(index_parent=True`) |
| ----------------------- | ----------------- | ---------------| ------ | 
| `f = load(...)`                |       7.4s                   |     0.6s  | 0.6s |
| `f = load(...); f.dm`       |       7.1s                   |      7.3s | 7.3s |
| `f = load(); f.halos()`    |       7.5s                   |    0.8s       | 0.8s | 
| `f = load(); h = f.halos(); h[1].properties`    |        29.6s   |    0.8s*   | 29.3s | 
 
*: h[1] cannot access its parent snapshot data, i.e. h[1]["pos"]  won't work, which allows for the massive speedup


pynbody/halo/__init__.py
```diff
@@ -25,6 +25,37 @@
 
 logger = logging.getLogger("pynbody.halo")
 
+class SimpleHalo(snapshot.ContainerWithPhysicalUnitsOption):
```
apontzen, line 28: I'm not sure I understand the difference between this and `DummyHalo`. Could it be explained somehow (is there perhaps a better name?)

pynbody/halo/adaptahop.py
```diff
@@ -71,7 +70,7 @@ class BaseAdaptaHOPCatalogue(HaloCatalogue):
     _halo_attributes_contam = tuple()
     _header_attributes = tuple()
 
-    def __init__(self, sim, fname=None, read_contamination=None, longint=None):
+    def __init__(self, sim, fname=None, read_contamination=None, longint=None, index_parent=True):
```
apontzen, line 73: Suggest adding a docstring to explain these various options

pynbody/halo/adaptahop.py
```diff
@@ -102,9 +102,7 @@ def __init__(self, sim, fname=None, read_contamination=None, longint=None):
         # Call parent class
         super(BaseAdaptaHOPCatalogue, self).__init__(sim)
 
-        # Initialize internal data
-        self._base_dm = sim.dm
-
+        self._sim = sim  # prevent garbage collection
```
apontzen, line 105: I suspect there is a reason why the original class stored `sim.dm` rather than `sim` itself. Specifically, `pynbody` recreates a new object every time `sim.dm` is called, which may be inefficient and/or lead to confusion (since generally `sim.dm is sim.dm` may evaluate as `False`!)

Possible that this doesn't actually matter here, but just wanted to check it had been thought through.

apontzen (commented):
Looks good, here are some small suggestions before merge

cphyc:
I realised there already is a similar mechanism for rockstar using a different approach. I will factorize it, but I'm switching the PR to draft in the meantime.

PR #635, opened 2021-12-10
Cache unit conversion (13× speedup in halo catalogue access)

cphyc:
When converting units from a halo catalogue, a *lot* of time is spent doing the `dimensional_project` operation. This PR adds support for a caching mechanism to speed things up.

| Action | Time before (s) | Time with 783033a (s) | Time with 3f4f171   (s) |
|---|--|--|--|
| `f = pynbody.load(…)`| 2 | 2 | 2 |
| `halos = f.halos(); halos.physical_units()` | 0.5 | 0.5 | 0.5 |
| `halos[1]`| 25 | 25 | 25 |
| `[h.properties for h in halos]` | ~2000 | ~300 | ~150 |

### DIFF
--- a/pynbody/snapshot/snapshot_util.py
+++ b/pynbody/snapshot/snapshot_util.py
@@ -29,7 +29,7 @@ class ContainerWithPhysicalUnitsOption:
         dims_s = repr(tuple(dims))
         key = (from_unit_s, dims_s, ucut)
         if key in cls._units_conversion_cache:
-            return cls._units_conversion_cache[key].copy()
+            return cls._units_conversion_cache[key]
 
         try:
             d = from_unit.dimensional_project(dims)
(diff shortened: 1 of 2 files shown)


pynbody/snapshot/snapshot_util.py
```diff
@@ -21,14 +21,42 @@ class ContainerWithPhysicalUnitsOption:
...
+    _units_conversion_cache = {}
+
+    @classmethod
+    def _cached_unit_conversion(cls, from_unit, dims, ucut=3):
+        key = (
+            repr(from_unit._bases),
+            repr(from_unit._powers),
+            tuple(dims),
+            ucut,
+        )
```
apontzen, line 33: Would it be better to have units implement a hash, to prevent needing to know internal structure of the class at this point?

pynbody/units.py
```diff
@@ -280,6 +280,8 @@ class NoUnit(UnitBase):
 
     def __init__(self):
         self._no_unit = True
+        self._bases = []
+        self._powers = []
```
apontzen, line 284: This again suggests to me a better way forward is a `__hash__` so that we don't need to worry about exposing internal state outside the classes themselves

apontzen (changes requested):
Nice work, I'd prefer to solve the hashing issue in a more pythonic way though, without having to rely on the internal state of the objects

pynbody/snapshot/snapshot_util.py
cphyc: I have implemented a `_hash_dimensions` method that returns a string unique to its _dimensions_, so that any two units having the same dimensions have the same hash.
```python
units.cm._hash_dimensions() == units.kpc._hash_dimensions() == units.Unit("3e8 m s**-1 yr")._hash_dimensions()
```

Issue #636, opened 2021-12-16
Issues in the calculation of the spherical polar coordinates


Martin-Rey:
Hi,

I think I have found two issues with the derived computations transforming into spherical polar coordinates:

1) The computation of the theta-component of the velocity in spherical polar coordinates is currently bugged: 
https://github.com/pynbody/pynbody/blob/5e97777552161c29eaed37f20252c5b815041e49/pynbody/derived.py#L108-L113

where the last velocity component should be `vz` rather than `vy`

Here are some minimal tests demonstrating the issue:
```
# Should find that x = r cos(azimuth) cos(polar declination)
#                  y = r sin(azimuth) cos(polar declination)
#                  z = r cos(polar declination)

np.testing.assert_allclose(sphere.st['x'], sphere.st['r'] * np.cos(sphere.st['az']) * np.sin(sphere.st['theta']))
np.testing.assert_allclose(sphere.st['y'], sphere.st['r'] * np.sin(sphere.st['az']) * np.sin(sphere.st['theta']))
np.testing.assert_allclose(sphere.st['z'], sphere.st['r'] * np.cos(sphere.st['theta']))

# Should find vr = vx * cos(azimuth) sin(polar declination) + vy sin(azimuth) sin(polar declination) + vz cos (polar declination)
vr = np.sin(sphere.st['theta']) * np.cos(sphere.st['az']) * sphere.st['vx'] + np.sin(sphere.st['theta']) * np.sin(sphere.st['az']) * sphere.st['vy'] + np.cos(sphere.st['theta']) * sphere.st['vz']
np.testing.assert_allclose(vr, sphere.st['vr'])

# Should find vphi = - vx * sin(azimuth) + vy cos(azimuth) + 0
vphi = - np.sin(sphere.st['az']) * sphere.st['vx'] + np.cos(sphere.st['az']) * sphere.st['vy']
np.testing.assert_allclose(vphi, sphere.st['vphi'])

# Should find vtheta = vx * cos(azimuth) cos(polar declination) + vy sin(azimuth) cos(polar declination) - vz sin (polar declination)
vtheta = np.cos(sphere.st['theta']) * np.cos(sphere.st['az']) * sphere.st['vx'] + np.cos(sphere.st['theta']) * np.sin(sphere.st['az']) * sphere.st['vy'] - np.sin(sphere.st['theta']) * sphere.st['vz']
np.testing.assert_allclose(vtheta, sphere.st['vtheta'])

[...]

PR #637, opened 2021-12-17
Fixing bug in theta velocity and adding tests on spherical derivations

Martin-Rey:
This PR partially address #636, fixing a bug in the derivation of the theta-velocity in spherical polar coordinates and adding a test for it.

### DIFF
--- a/pynbody/snapshot/__init__.py
+++ b/pynbody/snapshot/__init__.py
@@ -345,6 +345,13 @@ class SimSnap(object):
 
         if name in list(self.keys()):
             self._dependency_tracker.touching(name)
+
+            # Ensure that any underlying dependencies on 1D positions and velocities
+            # are forwarded to 3D dependencies as well
+            nd_name = self._array_name_1D_to_ND(name)
+            if nd_name is not None:
+                self._dependency_tracker.touching(nd_name)
+
             return self._get_array(name)
         
         with self._dependency_tracker.calculating(name):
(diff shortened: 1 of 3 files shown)


pynbody/snapshot/__init__.py
```diff
@@ -345,6 +345,13 @@ def _get_array_with_lazy_actions(self, name):
 
         if name in list(self.keys()):
             self._dependency_tracker.touching(name)
+
+            # Ensure that any underlying dependencies on 1D positions and velocities
+            # are forwarded to 3D dependencies as well
+            nd_name = self._array_name_1D_to_ND(name)
+            if nd_name is not None:
+                self._dependency_tracker.touching(nd_name)
```
apontzen, line 353: Is this right to duplicate the effort, or should this be _instead_ of `touching(name)` above?

Martin-Rey: I am not sure. I ended up coding it as an addition for two reasons: 

(i) my pre-designed tests for the behaviour I would expect were not passing when trying dependencies only on 3D names (your _instead_ solution). I didn't fully understand why, but some other arrays would then end up not being updated following coordinate transforms.

(ii) I agree the current solution (ensuring that 3D arrays have at least all dependencies of their 1D arrays)  is safer but requires slightly more effort. However since most positions arrays end up `touching` nearly every other (which is the reason why the code is currently almost perfectly working), the current solution only adds one or two additional checks to the existing ones, so it's not "duplicating" much work.

apontzen: OK, let's accept it as is as one day (haha) the whole snapshot class could do with an overhaul

PR #641, opened 2021-12-24
housekeeping fix for RAMSES working with AHF files

EthTay:
places AHF files inside snapshot folders rather than on the same level as them

assuming a file structure of "simulations"/"snapshots", with 5 or so AHF files per snapshot this previously would get very messy having many AHF files in the <simulations> folder, this just places them inside the <snapshot> folder instead for a cleaner implementation

### DIFF
--- a/pynbody/halo/ahf.py
+++ b/pynbody/halo/ahf.py
@@ -57,10 +57,13 @@ class AHFCatalogue(HaloCatalogue):
         if not self._can_load(sim, ahf_basename):
             self._run_ahf(sim)
 
-        HaloCatalogue.__init__(self,sim)
+        HaloCatalogue.__init__(self, sim)
 
         if use_iord is None:
-            use_iord = isinstance(sim.ancestor, (snapshot.gadget.GadgetSnap, snapshot.gadgethdf.GadgetHDFSnap))
+            use_iord = isinstance(
+                sim.ancestor,
+                (snapshot.gadget.GadgetSnap, snapshot.gadgethdf.GadgetHDFSnap)
+            )
 
         self._use_iord = use_iord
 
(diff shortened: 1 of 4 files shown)


pynbody/halo/ahf.py
```diff
@@ -66,8 +66,13 @@ def __init__(self, sim, make_grp=None, get_all_parts=None, use_iord=None, ahf_ba
         self._dosort = dosort
         self._only_stat = only_stat
 
+        RamsesTF = isinstance(sim.ancestor, (snapshot.ramses.RamsesSnap))
```
cphyc, line 69: Why is there anything special to do with RAMSES outputs? Is it because we need to convert the RAMSES output to tipsy/gadget output which is then fed into AHF?
If so, it may be cleaner to have a more generic solution for this, rather than having RAMSES-related code in the AHF frontends. I'd be happy to discuss this!

EthTay: Yeah thats exactly why this came about, but this is also because of the way you point to the ramses files when you load them in via pynbody you point to the level in which it expects the AHF files. For example with gadget/tipsy this isnt a problem because you point directly to the simulation files where as with ramses its a folder you point to, and so things could get very cluttered very fast

cphyc: I see. What do you think of the two suggestions I am making?

pynbody/halo/ahf.py
```diff
@@ -66,8 +66,13 @@ def __init__(self, sim, make_grp=None, get_all_parts=None, use_iord=None, ahf_ba
...
+
         if ahf_basename is not None:
             self._ahfBasename = ahf_basename
+        elif RamsesTF:
+            self._ahfBasename = util.cutgz(
+                glob.glob(sim._filename + '/' + '*z*AHF_halos*')[0])[:-5]
         else:
             self._ahfBasename = util.cutgz(
                 glob.glob(sim._filename + '*z*AHF_halos*')[0])[:-5]
```
cphyc, line 78:
```suggestion
        if ahf_basename is not None:
            self._ahfBasename = ahf_basename
        else:
            if os.path.isdir(sim._filename):
                glob_pattern = os.path.join(sim._filename, "*z*AHF_halos*")
            else:
                glob_pattern = f"{sim._filename}*z*AHF_halos*"
            self._ahfBasename = util.cutgz(
                glob.glob(glob_pattern)[0])[:-5]
            )
```

PR #642, opened 2022-01-20
Allow AdaptaHOP dataset to use only dummy halos

cphyc:
This allows for faster loading of the halo catalogue, but the
dummy halos only contain the halos' metadata (you cannot use
halo['pos'] to access the position of particles in the halo
for example).


cphyc:
pre-commit.ci autofix

PR #643, opened 2022-01-26
BUGFIX: Correctly iterate one-indexed halo catalogues

cphyc:
This fixes a bug where the last halo in a halo catalogue would not be listed when doing
```python
halos = f.halos()
list(halos)  # is missing the last halo
```
This only happens with one-indexed halo catalogues (e.g. AdaptaHOP).


apontzen:
Thanks!

PR #646, opened 2022-02-01
Fix selecting SUBFIND haloes with precalculated indices

j-davies-astro:
This is quite a substantial change to a very integral part of pynbody, but I think it's an important fix for those using pynbody and SUBFIND together, particularly when also using tangos. I've been using pynbody with this change in place for many months and haven't encountered an issue, but happy to iterate on it a bit. 

With pynbody, one can sort a SimSnap by the unique ids assigned by the halo finder (or by pynbody itself), and run the precalculate() routine to find the "start and stop" indices of each halo, speeding up the selection of multiple haloes from that SimSnap. This process is fundamental to the functioning of the tangos module.

Currently, this process assumes that there are exactly as many of these unique indices as there are haloes in the simulation, running from 1 to N_haloes. Problems arise, however, when a placeholder id is used to signify e.g. unbound particles. This is the case for EAGLE's implementation of SUBFIND, as an example - unbound particles are assigned the id 2^30. In this case, _boundaries is a length 2^30 array containing mostly -1's between the index of the final "real" halo in the box and the final element. Not only does this waste a lot of memory, but the "stop" index of the final halo is assumed to be the next element in _boundaries, which will be -1. This causes all (but one) of the unbound particles in the SimSnap to be assigned to the final halo, which is clearly problematic - it tends to lead to the creation of one big "super halo" in tangos made up of all the unbound particles in the simulation.

To resolve this, I switched from using "util.find_boundaries" to using a combination of np.unique and np.searchsorted to identify all the unique halo finder IDs in the simulation, and identify the boundaries of each halo. With these unique IDs (_unique_i) in hand, the boundaries of halo i can be easily found by searching for i in _unique_i.

As far as I can tell, this method is considerably faster and more memory-efficient for simulations employing subfind (my tangos runs are sped up by factors of 5-6ish), and hopefully shouldn't cause any issues for other simulations.

### DIFF
--- a/pynbody/halo/__init__.py
+++ b/pynbody/halo/__init__.py
@@ -296,8 +296,8 @@ class GrpCatalogue(HaloCatalogue):
         getting a single halo, however."""
         self._sorted = np.argsort(
             self.base[self._array], kind='mergesort')  # mergesort for stability
-        self._boundaries = util.find_boundaries(
-            self.base[self._array][self._sorted])
+        self._unique_i = np.unique(self.base[self._array])
+        self._boundaries = np.searchsorted(self.base[self._array][self._sorted],self._unique_i)
 
     def get_group_array(self, family=None):
         if family is not None:
@@ -317,20 +317,18 @@ class GrpCatalogue(HaloCatalogue):
             return index
         else:
             # pre-calculated
-            if i >= len(self._boundaries) or i < 0:
-                raise no_exist
-            if self._boundaries[i] < 0:
+            if not np.isin(i,self._unique_i):
                 raise no_exist
 
-            start = self._boundaries[i]
-            if start is None:
-                raise no_exist
+            match = np.where(self._unique_i==i)[0]
 
-            end = None
-            j = i + 1
-            while j < len(self._boundaries) and end is None:
-                end = self._boundaries[j]
-                j += 1
+            start = self._boundaries[match][0]
+            
+            if start == self._boundaries[-1]:
+                # This is the final halo
+                end = None
+            else:
+                end = self._boundaries[match+1][0]
 
             return self._sorted[start:end]
 


apontzen:
I agree this implementation looks fine. However it would be very good to add a test which ensures the correct behaviour with a catalogue of this sort. I can add a small simulation/subfind catalogue to the testdata if you have one?

j-davies-astro:
Please ignore those last two commits - nothing has changed, I'm just amending the author of the commit as it was set up incorrectly.

I have actually uncovered something odd - my changes here still don't actually prevent all the unbound particles from being dumped into one "super halo", but for some reason, they stop tangos from interpreting this object as the "first halo" in the simulation. I'm still not sure why this is. I'm going to make a further change that will prevent pynbody from interpreting this spurious "super halo" as a real halo and push it shortly.

As for the test simulation, I have added "TangosSubGroupNumber" to an EAGLE 12 Mpc 188^3 volume, which we could use for testing? Or we could just use one of the MW GM runs?

j-davies-astro:
OK, I've done some digging and am a little perplexed because I now can't see why this problem originally arose when using tangos with this precalculation method. While a lot faster, the new method proposed here still leaves you with a single "super-halo" made up of all the unbound particles. Tangos is already able to deal with this, however, as when the group catalogue is made by the EagleLikeInputHandler (in tangos/input_handlers/eagle) it instructs pynbody to ignore the final halo. This object still makes it into the halo catalogue, but len(halo_catalogue) ignores it, so when tangos loops from halo 1 to halo len(halo_catalogue) + 1, this spurious object is ignored. The only way the spurious halo would be included is if there were 2^30-1 haloes in the box, which seems unlikely!

This all leaves me none the wiser as to why this was a problem for DM-only simulations in the first place. I suspect this may have arisen from tangos identifying EAGLE outputs by looking for PartType0, which doesn't exist in DM-only outputs, and therefore defaulting to the vanilla PynbodyInputHandler, which doesn't ignore the final halo like this. One of my other recent pull requests fixed this issue for pynbody, but I can't quite work out how tangos decides which input handler to use. At any rate, this doesn't seem to be a problem any more and tangos always interprets DM-only Eagle outputs correctly.

I still think this change to the precalculation is worth implementing though, as it seems to be considerably faster (at least for my simulations). I'm not sure whether I should delete the spurious final halo before performing the precalculation routine (as I suggested above) as users might want to perform precalculation and then identify all unbound particles - though this seems a very roundabout way of doing something very straightforward.

As for performing a test, I don't think these snapshots are any different from standard non-subfind ones in which all haloes have a unique id, so does this need another test? What might be necessary is to test that haloes found after performing precalculation are the same as those found without it - I don't think such a test exists yet?

apontzen:
Discussed and realised there is no test to reveal the changed behaviour since externally the behaviour is identical.

PR #647, opened 2022-02-01
Fixed issue with number of vectors in velocity image

j-davies-astro:
Minor fix for a small bug in the velocity_image function that I came across. Currently, the coordinates of the velocity image are defined using np.arange, which could sometimes lead to the creation of an array that is 1 element too long in each dimension.

For example, by default pynbody creates a 40x40 image of the velocity field, and then combines it with X, Y (defined using np.meshgrid) to produce the arrows on the image. Using np.arange can lead to a (41,41) shape array being created, which is incompatible with the velocity image.

Using np.linspace instead means you're guaranteed to get X and Y of correct size.

### DIFF
--- a/pynbody/plot/sph.py
+++ b/pynbody/plot/sph.py
@@ -111,11 +111,11 @@ def velocity_image(sim, width="10 kpc", vector_color='black', edgecolor='black',
             width = _units.Unit(width)
         width = width.in_units(sim['pos'].units, **sim.conversion_context())
 
-    width = float(width)
+    width = np.float32(width)
 
-    pixel_size = width / vector_resolution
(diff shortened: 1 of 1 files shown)


pynbody/plot/sph.py
```diff
@@ -111,11 +111,11 @@ def velocity_image(sim, width="10 kpc", vector_color='black', edgecolor='black',
             width = _units.Unit(width)
         width = width.in_units(sim['pos'].units, **sim.conversion_context())
 
-    width = float(width)
+    width = np.float32(width)
```
apontzen, line 114: Why should this specifically be `np.float32` rather than just `float`? I don't quite get it. (Similar point on the line below)

apontzen:
I agree with the fix but confused by the introduction of `np.float32`

PR #648, opened 2022-02-01
Added ability to select between different SPH smoothing kernels

j-davies-astro:
I've added an extra option to the sph routines that allows the user to pick a different kernel for making images. New kernels need to be defined as functions in the Kernel() class, and so far I've only added 'spline' (the current pynbody default) and 'wendlandC2', which is the kernel EAGLE uses. If no kernel is specified, 'spline' is used.


pynbody/sph/__init__.py
```diff
@@ -204,7 +206,23 @@ def get_samples(self, dtype=np.float32):
 
     def get_value(self, d, h=1):
         """Get the value of the kernel for a given smoothing length."""
-        # Default : spline kernel
+
+        if self.type == 'spline':
+            # Default : spline kernel
+            return self.spline(d,h)
+
+        elif self.type == 'wendlandC2':
```
apontzen, line 214: This should be implemented as a subclass, not as a string option

pynbody/sph/__init__.py
```diff
@@ -214,13 +232,23 @@ def get_value(self, d, h=1):
...
+        else:
+            f = 0
+
+        return (21. * f) / (16. * math.pi * h ** 3)
+
 
 class Kernel2D(Kernel):
 
-    def __init__(self, k_orig=Kernel()):
+    def __init__(self, type='spline'):
```
apontzen, line 248: The parameter passed here should indeed be a `Kernel` object, and an appropriate subclass can then be passed if one wants to use a Wendland kernel

pynbody/plot/sph.py
```diff
@@ -364,7 +368,7 @@ def image(sim, qty='rho', width="10 kpc", resolution=500, units=None, log=True,
 
     width = float(width)
 
-    kernel = sph.Kernel()
+    kernel = sph.Kernel(type=kernel_type)
```
apontzen, line 371: Here you can create the appropraite kernel object based on the string passed in (i.e. it's ok for the user-facing bit to have strings)

pynbody/plot/sph.py
```diff
@@ -375,14 +379,14 @@ def image(sim, qty='rho', width="10 kpc", resolution=500, units=None, log=True,
         is_projected = _units_imply_projection(sim, qty, units)
 
     if is_projected:
-        kernel = sph.Kernel2D()
+        kernel = sph.Kernel2D(type=kernel_type)
```
apontzen, line 382: This line should read `kernel = sph.Kernel2D(kernel)`, to construct a 2D kernel over whatever kernel has been previously selected

pynbody/plot/sph.py
```diff
@@ -375,14 +379,14 @@ def image(sim, qty='rho', width="10 kpc", resolution=500, units=None, log=True,
...
 
     if av_z:
         if isinstance(kernel, sph.Kernel2D):
             raise _units.UnitsException(
                 "Units already imply projected image; can't also average over line-of-sight!")
         else:
-            kernel = sph.Kernel2D()
+            kernel = sph.Kernel2D(type=kernel_type)
```
apontzen, line 389: see comment above

pynbody/sph/__init__.py
```diff
@@ -181,7 +181,7 @@ def rho(self):
 
 class Kernel(object):
 
-    def __init__(self):
+    def __init__(self,type='spline'):
```
apontzen, line 184: This option should not be present, rather a subclass should be implemented for a different kernel type

apontzen (changes requested):
Nice to do this but the change in the construction of Kernels will break some other projects. The Wendland kernel should be implemented as a new subclass i.e.

```python
class WendlendKernel(Kernel):
    def get_value(...):
        ....
```

That is the basic point, some detailed comments inline

pynbody/plot/sph.py
```diff
@@ -364,25 +368,31 @@ def image(sim, qty='rho', width="10 kpc", resolution=500, units=None, log=True,
 
     width = float(width)
 
-    kernel = sph.Kernel()
+    if kernel_type == 'wendlandC2':
+        kernel = sph.WendlandC2Kernel()
+    elif kernel_type == 'spline':
+        kernel = sph.Kernel()
+    else:
+        print('Invalid kernel_type specified (options are "spline" and "wendlandC2"). Defaulting to a cubic spline...')
```
apontzen, line 376: This should be `raise ValueError` not `print`

j-davies-astro:
Thanks for the feedback, I think the implementation should be correct now. Happy to change the way it's handled if the user requests an invalid kernel - right now it defaults back to spline with a printed warning but if you'd like to put something more fancy in I can make further changes.

pynbody/plot/sph.py
```diff
@@ -339,6 +340,9 @@ def image(sim, qty='rho', width="10 kpc", resolution=500, units=None, log=True,
     *linthresh* (None): if the image has negative and positive values
      and a log scaling is requested, the part between `-linthresh` and
      `linthresh` is shown on a linear scale to avoid divergence at 0
+
+    *kernel_type* ('spline'): SPH kernel to use for smoothing. Defualts to a
```
apontzen, line 344: typo (Def*ua*lts)

apontzen (commented):
Almost there!

pynbody/plot/sph.py
j-davies-astro: whoops!

apontzen:
Sorry one last thing - I think it's worth checking this actually works (and continues to work) with a test. Under `sph_image_test` you will find a `test_images` function which you could easily adapt to also check asking for the alternative kernel behaves as expected (ie. presumably it should return a different image)

apontzen:
Sorry I meant the latter - check the image is not only different but also correct. Ie just add one more reference numpy file to test against.

apontzen (approved):

PR #649, opened 2022-02-01
Support reading hydro file descriptor blocks

cphyc:
RAMSES ships with so-called `file_descriptor`s that describe what variables are written on-disk.

This adds support for reading them automatically.


apontzen:
Can we test this in some way?

cphyc:
@apontzen I have added a test (see 064fa3f).

PR #650, opened 2022-02-02
Convert conformal time

cphyc:
RAMSES stores the formation time of objects using conformal time, which then need to be converted to physical times for analysis.

At the moment, the conversion is made using an external tool (`part2birth`, that is shipped along with RAMSES). However, this is quite brittle and requires creating a large number of sidecar files (as many as the number of processors used to run the simulation).

This PR simplifies the process by converting the conformal time to physical time directly from Python without the need for an external tool. Note that the values obtained thus may slightly differ from those obtained with `part2birth` at the percent level.

In addition, I have checked that the results agree within 1% or less in a large cosmological simulation (made of 7,143,831 stellar particles) with the following code:
```python
import pynbody

f = pynbody.load("../data/halo_000480/hydro_reference/output_00200")
f.star["tform_python"] = pynbody.analysis.ramses_util.get_tform(f.star, use_part2birth=False)
f.star["tform_ramses"] = pynbody.analysis.ramses_util.get_tform(f.star, use_part2birth=True)

np.testing.assert_allclose(
    f.star["tform_python"],
    f.star["tform_ramses"],
    rtol=5e-3,
)
```

### DIFF
--- a/pynbody/analysis/_cosmology_time.pyx
+++ b/pynbody/analysis/_cosmology_time.pyx
@@ -1,0 +1,83 @@
+cimport numpy as np
+
+import numpy as np
+
+from libc.math cimport sqrt
+
+
+cdef double dadtau(double aexp_tau,double O_mat_0,double O_vac_0,double O_k_0):
+    cdef double aexp_tau3 = aexp_tau * aexp_tau * aexp_tau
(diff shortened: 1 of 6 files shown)


earnric:
Great stuff! Thanks!

PR #653, opened 2022-03-11
Modernize with pyupgrade

cphyc:
This uses `pyupgrade --py3-plus`  on all .py files


apontzen (commented):
Thanks Corentin. In general I am a bit nervous of automated code transformation tools. Is this tool well tested?

That said, it's good to do this clean-up and I don't see anything it's done that would obviously be a problem. I've left a few minor comments, given that we are doing some cleaning it would be good to look at these before merging.

cphyc:
Hi @apontzen, the pyupgrade package has been in use in the yt project for a year now (as a pre-commit hook) with zero issue so far!

PR #654, opened 2022-03-14
Dot not use bare excepts

cphyc:
This addresses the flake8 recommendation of not using bare except that catch any exception, including `KeyboardInterrupt` (see https://www.flake8rules.com/rules/E722.html).


cphyc:
pre-commit.ci autofix

cphyc:
Note: this PR is going to fail until #660 is merged in.

PR #661, opened 2022-03-24
Support RAMSES/AHF combination

cphyc:
This is built on top of #641.


EthTay (commented):
line 443 should actually not have the +1 and should read as:

`halo_index = ID2index[haloid]`

but also compared to my iteration of this, this is definitely better and fixes a few problems i have been having

apontzen:
@cphyc ping about Ethan's comment above - do you agree/are you able to fix?

EthTay:
in addition to the above change there are two extra changes which is needed being:

line 429 should be possibly removed as there is no crazy number, at least in build 94 what this line actually reads is the haloid and number of sub halos which when removed is correctly read by line 439

and

line 444-446 changed to `children = [  ID2index[  int(x)  ] for x in f.readline().split()   ]`

which is again just the removal of the +1 to int(x) when used with the inital change in my other comment this then correctly maps the subhalo IDs into their parent when calling h[1].properties['children'] for example

EthTay:
sorry ive caught this at a bad time, but theres one change which is on line 440 which currently reads:

` halo_index = ID2index[haloid + 1]`

this should also be updated to be

`halo_index = ID2index[haloid]`

sorry i didnt spot this i thought i included all the changes i made on this, but ive just tested this in my own repo and this should be it now, the +1 is what is currently breaking things @apontzen @cphyc

apontzen:
Oh dear - could you provide a new pull request to fix this up? Thanks!

EthTay:
just submitted the pr now #681 should be this final touch, ill also close my other pr that this was built up from as well

PR #667, opened 2022-05-27
Build wheel on M1 macs

apontzen:
This probably won’t work as will likely need to switch compilers, but it’s a first attempt


apontzen:
Too hard to make this work unless/until GitHub provide arm64 macOS runners

PR #668, opened 2022-05-29
Simplify/modernise installation

apontzen:
Switch cython to language level 3
Remove OpenMP detection (just let it fail with Apple's clang, since it's very hard to keep it working)


apontzen:
pre-commit.ci autofix

PR #670, opened 2022-06-03
Build linux-aarch64 wheel

apontzen:
Build aarch64 using qemu on linux


apontzen:
pre-commit.ci autofix

PR #672, opened 2022-06-14
Reading units issue solved

thanatom:
Cosmological factors a and h were impossible to retrieve from AREPO snapshots and the code, consequently, crashed.

In the function _get_cosmo_factors, the case where the attributes 'aexp-scale-exponent', 'aexp-scale-exponent', 'aexp-scale-exponent', and ''aexp-scale-exponent'' was not handled.

I forced gadget-like factors in this case.

### DIFF
--- a/pynbody/snapshot/gadgethdf.py
+++ b/pynbody/snapshot/gadgethdf.py
@@ -563,9 +563,18 @@ class GadgetHDFSnap(SimSnap):
 
         cosmo = 'HubbleParam' in list(self._get_hdf_parameter_attrs().keys())
         if cosmo:
-            for fac in self._get_cosmo_factors(self._hdf_files[0], 'Coordinates'): dist_unit *= fac
-            for fac in self._get_cosmo_factors(self._hdf_files[0], 'Velocities'): vel_unit *= fac
-            for fac in self._get_cosmo_factors(self._hdf_files[0], 'Mass'): mass_unit *= fac
+            try:
+                for fac in self._get_cosmo_factors(self._hdf_files[0], 'Coordinates'): dist_unit *= fac
+            except:
+                dist_unit *= units.a * units.h**-1
+            try:
+                for fac in self._get_cosmo_factors(self._hdf_files[0], 'Velocities'): vel_unit *= fac
+            except:
+                vel_unit *= units.a**(1,2)
+            try:
+                for fac in self._get_cosmo_factors(self._hdf_files[0], 'Mass'): mass_unit *= fac
+            except:
+                mass_unit *= units.h**-1
 
         self._file_units_system = [units.Unit(x) for x in [
             vel_unit, dist_unit, mass_unit, "K"]]
--- a/pynbody/snapshot/gadgethdf.py
+++ b/pynbody/snapshot/gadgethdf.py
@@ -365,9 +365,10 @@ class GadgetHDFSnap(SimSnap):
     def _get_cosmo_factors(hdf, arr_name) :
         """Return the cosmological factors for a given array"""
         match = [s for s in GadgetHDFSnap._get_hdf_allarray_keys(hdf) if ((s.endswith("/"+arr_name)) & ('PartType' in s))]
-        if len(match) == 0:
-                # no match
-            return False, False
+        if (arr_name == 'Mass' or 'Masses') and len(match) == 0:
+            # mass stored in header. We're out in the cold on our own.
+            warnings.warn("Masses are either stored in the header or have another dataset name; assuming the cosmological factor %s" % units.h**-1)
+            return units.Unit('1.0'), units.h**-1
         if len(match) > 0 :
             try:
                 aexp = hdf[match[0]].attrs['aexp-scale-exponent']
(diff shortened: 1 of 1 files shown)


pynbody/snapshot/gadgethdf.py
```diff
@@ -563,9 +563,18 @@ def __init_unit_information(self):
 
         cosmo = 'HubbleParam' in list(self._get_hdf_parameter_attrs().keys())
         if cosmo:
-            for fac in self._get_cosmo_factors(self._hdf_files[0], 'Coordinates'): dist_unit *= fac
-            for fac in self._get_cosmo_factors(self._hdf_files[0], 'Velocities'): vel_unit *= fac
-            for fac in self._get_cosmo_factors(self._hdf_files[0], 'Mass'): mass_unit *= fac
+            try:
+                for fac in self._get_cosmo_factors(self._hdf_files[0], 'Coordinates'): dist_unit *= fac
+            except:
```
apontzen, line 568: Please change this bare except to a more specific exception e.g. `except KeyError`, if that's the appropriate one

pynbody/snapshot/gadgethdf.py
```diff
@@ -563,9 +563,18 @@ def __init_unit_information(self):
...
+                dist_unit *= units.a * units.h**-1
```
apontzen, line 569: I think it would be appropriate to raise a warning here e.g.

`warnings.warn("Unable to find cosmological factors in HDF file; assuming position is %s"%dist_unit)`

or something along those lines

apontzen (commented):
Thanks! This looks good but here are some suggestions to polish it off

pynbody/snapshot/gadgethdf.py
thanatom: I agree to update all the excepts.

pynbody/snapshot/gadgethdf.py
thanatom: I'll add all the warning messages.

pynbody/snapshot/gadgethdf.py
```diff
@@ -365,9 +365,9 @@ def _get_hdf_dataset(self, particle_group, hdf_name):
     def _get_cosmo_factors(hdf, arr_name) :
         """Return the cosmological factors for a given array"""
         match = [s for s in GadgetHDFSnap._get_hdf_allarray_keys(hdf) if ((s.endswith("/"+arr_name)) & ('PartType' in s))]
-        if arr_name == 'Mass' and len(match) == 0:
-            # mass stored in header. We're out in the cold on our own.
-            return units.Unit('1.0'), units.h**-1
+        if len(match) == 0:
+                # no match
+            return False, False
```
apontzen, line 370: This could break other code which relies on `_get_cosmo_factors` (it is called from several places). I think it would be better to revert this but issue a warning from here to match the type of warning you are issuing below.

apontzen (commented):
Thanks, almost there but I am concerned about changing the returns from `_get_cosmo_factors` -- see comments

pynbody/snapshot/gadgethdf.py
thanatom: Since the mass dataset could be named both "Mass" and "Masses", this "if" statement was triggered in both the following cases:
1- If the mass dataset is missing and masses are stored in the header (as, I assume, it was originally conceived)
2- If the mass dataset is present but has a different name.

As a consequence, in __init_unit_information(), if one wants to check for the presence, e.g., of both "Mass" and "Masses" a factor h^-2 is accumulated if the case where "no match is found" is not handled.
I opted to modify _get_cosmo_factors() in order to disentangle these two cases and also because I did not find any other usage of it in the whole folder.

apontzen: I'm not fully following, but I agree the current logic is rather tortured. My point is different -- I don't think it ever makes sense to return a tuple of `False, False`. This could cause huge confusion to callers. If the condition should never be encountered, it should raise an exception. If the condition can lead to a default guess, that default should be returned and a warning issued.

You are right it isn't called from elsewhere, I misread that -- but still, to enable its potential reuse and save caller confusion, I think the function should either return units, or raise an exception, but never return a boolean.

apontzen (commented):
The failing tests on python 3.8+ reflect that this is based off an old version of pynbody. It would be helpful for you to merge in the current master to check everything works well together.

I also spotted what I think is a typo/bug in the latest changes (see line comment below). 

Otherwise I think this is all good to go! Thanks for your work and I'll merge as soon as this final issue is ironed out

pynbody/snapshot/gadgethdf.py
```diff
@@ -365,8 +365,9 @@ def _get_hdf_dataset(self, particle_group, hdf_name):
     def _get_cosmo_factors(hdf, arr_name) :
         """Return the cosmological factors for a given array"""
         match = [s for s in GadgetHDFSnap._get_hdf_allarray_keys(hdf) if ((s.endswith("/"+arr_name)) & ('PartType' in s))]
-        if arr_name == 'Mass' and len(match) == 0:
+        if (arr_name == 'Mass' or 'Masses') and len(match) == 0:
```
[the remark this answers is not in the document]
apontzen: I am fine with the intent of the change ("masses" as well as "mass") - it's just the line as implemented has a bug in it. My suggestion would (I believe) fix it.

Regarding versioning, the issue is that you are so far behind the pynbody master branch. But this isn't a major issue, I can deal with that.

PR #673, opened 2022-06-21
Unify the Subfind implementation

apontzen:
The Subfind-HDF implementation for Gadget4/Arepo could only load one particle type at a time, which
is quite restrictive. Meanwhile, a more powerful Subfind-HDF implementation existed but was not
compatible with the changes made for Gadget4/Arepo. This expands the unified implementation to
work with Gadget4/Arepo.


apontzen:
pre-commit.ci autofix

PR #681, opened 2022-08-11
Update ahf.py

EthTay:
final ahf fix relating to PR #661


cphyc:
The tests failure seem to be independent of the PR. At least, I cannot reproduce the error locally. Would it be possible to retrigger them manually @apontzen? At

apontzen:
The test failures appear to be to do with a change in behaviour of pip. I am trying to solve them in a separate PR.

PR #688, opened 2022-11-15
Small fixes when dealing with AHF files

Martin-Rey:
- Fix 1: The detection of AHF catalogues currently cannot deal robustly with directory outputs (rather than files), as is typically the case for Ramses outputs. In such case, and if a trailing "/" is added at the end of the output, the method references two duplicate, possible catalogues, which then breaks the `_can_load` test which tests for a `len==1`. This PR  prevents this by removing the possibility to generate duplicate candidates, and raises more explicit error messages if no or multiple AHF files are found.

- Fix 2: AHF-generated properties can sometimes come out as `inf` from the catalogues, breaking the current enumeration which only checks for `NaN`

Let me know if anything is unclear!


Martin-Rey:
Sorry for the mess, I need pushing extra changes on this branch that I am currently working on. 

I think it's easier if the later changes and additions, which are quite orthogonal to this, are part of their own branches.

Martin-Rey:
I have now added a test of the behaviour, which relies on having a Ramses output with AHF catalogues (same as #697). The tests should fix themselves once we add this to the testdata.

PR #692, opened 2022-12-06
added support for different smoothing kernels in kdtree

Evafranck:
We have added a config option to choose the smoothing kernel that is applied in the calculation of smoothed properties in the kdtree. New option now is the Wendland C2 Kernel which can be set via a config option in the config file while the fiducial option is still set to the cubic spline kernel for back compatibility reasons.

Note: changing the config option in an interactive session only applies the new kernel if the simulation is reloaded.


pynbody/sph/kd.h
```diff
@@ -1,7 +1,7 @@
 #ifndef KD_HINCLUDED
 #define KD_HINCLUDED
 
-#include <Python.h>
+#include <python3.8/Python.h>
```
apontzen, line 4: This is causing builds to fail because it assumes Python 3.8

pynbody/sph/kdtree.py
```diff
@@ -157,15 +157,17 @@ def smooth_operation_to_id(self, name):
         else:
             raise ValueError("Unknown smoothing request %s" % name)
 
-    def populate(self, mode, nn):
+    def populate(self, mode, nn, Kernel = 'CubicSpline'):
```
apontzen, line 160: `Kernel` should be `kernel` -- lower case as per PEP8

pynbody/sph/smooth.cpp
```diff
@@ -654,18 +689,20 @@ void smDensity(SMX smx,int pi,int nSmooth,int *pList,float *fList)
 	for (j=0;j<nSmooth;++j) {
 		pj = pList[j];
 		r2 = fList[j]*ih2;
-		rs = 2.0 - sqrt(r2);
-		if (r2 < 1.0) rs = (1.0 - 0.75*rs*r2);
-		else rs = 0.25*rs*rs*rs;
-		if(rs<0) rs=0;
+		if (Wendland) {
```
apontzen, line 692: Does this branch inside a critical part of the code have performance implications? If it's only a small percentage, that's fine, but if it makes a big difference to smoothing runtimes I'd prefer to find a different way (e.g. using templates to generate independent code for the different kernels). Have you checked any runtimes?

apontzen (changes requested):
This is a really helpful patch which I am sure lots of people will appreciate. Thank you!

I have left some comments inline -- could you let me know what you think?

pynbody/sph/smooth.cpp
TobiBu: Hi Andrew,
we haven't checked for any runtime performance. This if statement was simply the most straight forward way of checking at runtime which kernel is requested. we are happy to rewrite this using templates but I am not entirely sure what's the best strategy here. shall we define a template for smDensity or a general template for Kernel depending on what's specified by a parameter?

apontzen: It may be that it is having little impact. In the first instance, I'd recommend just measuring the runtime for calculating the density across a large simulation before and after this patch. If that doesn't show a big change, no need for further action.

If it does show a big change, we want the compiler to optimise out the branch, and I can make more specific suggestions if they would be helpful.

pynbody/sph/kd.h
Evafranck: Thanks, fixed it.

pynbody/sph/smooth.cpp
TobiBu: so, testing on a simulation with about ~1.75e7 particles does not show a timing difference (2 runs: 44s to 43s and 38s to 36s). I guess we can proceed with the code as is for now. However, we realised that changing the config parameter during an interactive session will not trigger recalculation see e.g. density [in sph.__init__.py](https://github.com/pynbody/pynbody/blob/aa4bd2a0e9fb411d68c246daac7a2752e0020d2f/pynbody/sph/__init__.py#L159). I wonder if we should at least add a check somewhere or do something like for the [smoothing length](https://github.com/pynbody/pynbody/blob/aa4bd2a0e9fb411d68c246daac7a2752e0020d2f/pynbody/sph/__init__.py#L146). Not sure what the best practice here would be.

apontzen:
Sorry - had failed to spot the updates on this branch. Thanks for addressing my comments.

At the moment the tests are failing but I think this is because of the changes to numpy. These have been addressed on the latest pynbody main branch, so please could you merge that into here? Then we can see where things stand...

GFG-CHAOS:
Hi Andrew,

done. I guess we are good to go. Only the pre-commit trailing whitespace check fails now.

apontzen:
pre-commit.ci autofix

apontzen:
Did you want to do anything about the point you raised on recalculation?

Personally I think it will be a rare situation that somebody updates the kernel at runtime, and it's ok to let them 'fend for themselves' and work out how to recalculate it.

apontzen (approved):

GFG-CHAOS:
yes, that was also what we thought. Most use cases will set the kernel before analysis and not during a session.
I simply wanted to cross-check if there would be an easy fix to it or if that should be noted somewhere so that people are at least aware of it.

apontzen:
Thanks for this very nice piece of work!

PR #694, opened 2022-12-09
Fix kernel gradient

### LINKED ISSUE #693: wrong kernel gradient in smooth.cpp
Hi Andrew,

while working on PR #692 I realised that the kernel gradient of the cubic spline kernel in e.g. smooth.cpp in line 763 following is wrong.

This is what's in the code:
`r2 = fList[j];
q2 = r2*ih2;
r = sqrt(r2);
q = sqrt(q2);
// Kernel gradient
if (q < 1.0) rs = -3.0*ih + 2.25*r*ih2;
else rs = -0.75*(2-q)*(2-q)/r;
rs *= fNorm;`

While this is what I derive and what's also coded in e.g. [pySPH](https://pysph.readthedocs.io/en/latest/_modules/pysph/base/kernels.html#CubicSpline)

Can you confirm that 

1. one line needs to read: `if (q < 1.0) rs = -3.0*q + 2.25*q2;`
2. the other line simply needs to be: `else rs = -0.75*(2-q)*(2-q);`

If this is correct I'll go about and fix it in PR #692

TobiBu:
This fixes an incorrect calculation of the kernel gradient of the cubic spline kernel in smooth.cpp and updates the test file for the velocity curl and divergence which relies on the kernel gradient. This fix should close #693


TobiBu:
oh, wait with this. I might have confused myself. I'll let you know when I checked things again.

PR #695, opened 2022-12-15
Fix Ramses loader to only attempt to access metal and tform fields if they are actually on disc

### LINKED ISSUE #689: Ramses loading is currently unsafe if specific name are not present in its particle blocks
Hi

The current implementation of Ramses loading implicitly assumes that specific particle field names are always present on disc, raising a `ValueError` that is not caught by the loading process and blocking any subsequent attempt at deriving them. 

For example, my simulation is storing explicit metal arrays from individual elements (here Fe and O)
``` 
Particle fields
pos vel mass iord level tform metal_Fe massform kick metal_O
```
which are available to load in the list of `self._particle_blocks`

I have then defined a derived array for the total metallicity, called `metal` (basically doing `metal_Fe + metal_O`) which I then want to use with pynbody's plotting and SSP capabilities. This works perfectly for gas:
```
In [2]: s.g['metal']
Out[2]: 
SimArray([0.00000000e+000, 0.00000000e+000, 0.00000000e+000, ...,
          2.50245581e-253, 1.63412234e-253, 3.59176535e-254], '1.00e+00')
```

But fails for stars and DM with 
```
In [3]: s.st['metal']
ValueError                                Traceback (most recent call last)
<ipython-input-3-5b4ca7984d3f> in <module>
----> 1 s.st['metal']

/scratch/dp191/shared/python/anaconda3/lib/python3.8/site-packages/pynbody-1.2.3-py3.8-linux-x86_64.egg/pynbody/snapshot/__init__.py in __getitem__(self, i)
    256 
    257         if isinstance(i, str):
--> 258             return self._get_array_with_lazy_actions(i)
    259         elif isinstance(i, slice):
    260             return SubSnap(self, i)

/scratch/dp191/shared/python/anaconda3/lib/python3.8/site-packages/pynbody-1.2.3-py3.8-linux-x86_64.egg/pynbody/snapshot/__init__.py in _get_array_with_lazy_actions(self, name)
    354             if not self.lazy_off:
    355                 if not self.lazy_load_off:
--> 356                     self.__load_if_required(name)
    357                 if not self.lazy_derive_off:
    358                     self.__derive_if_required(name)

/scratch/dp191/shared/python/anaconda3/lib/python3.8/site-packages/pynbody-1.2.3-py3.8-linux-x86_64.egg/pynbody/snapshot/__init__.py in __load_if_required(self, name)
    365         if name not in list(self.keys()):
    366             try:
--> 367                 self.__load_array_and_perform_postprocessing(name)
    368             except OSError:
    369                 pass

/scratch/dp191/shared/python/anaconda3/lib/python3.8/site-packages/pynbody-1.2.3-py3.8-linux-x86_64.egg/pynbody/snapshot/__init__.py in __load_array_and_perform_postprocessing(self, array_name, fam)
    847             else:
    848                 try:
--> 849                     self._load_array(array_name, fam)
    850                 except OSError:
    851                     for fam_x in self.families():

/scratch/dp191/shared/python/anaconda3/lib/python3.8/site-packages/pynbody-1.2.3-py3.8-linux-x86_64.egg/pynbody/snapshot/__init__.py in _load_array(self, array_name, fam, **kwargs)
   1942     def _load_array(self, array_name, fam=None, **kwargs):
   1943         if fam is self._unifamily or fam is None:
-> 1944             self.base._load_array(array_name, self._unifamily)
   1945 
   1946     def _derive_array(self, array_name, fam=None):

/scratch/dp191/shared/python/anaconda3/lib/python3.8/site-packages/pynbody-1.2.3-py3.8-linux-x86_64.egg/pynbody/snapshot/ramses.py in _load_array(self, array_name, fam)
   1074 
   1075             elif array_name == 'metals' or array_name == 'metal':
-> 1076                 self._load_particle_block('metal')
   1077                 self._convert_metal_name()
   1078 

/scratch/dp191/shared/python/anaconda3/lib/python3.8/site-packages/pynbody-1.2.3-py3.8-linux-x86_64.egg/pynbody/snapshot/ramses.py in _load_particle_block(self, blockname)
    933 
    934     def _load_particle_block(self, blockname):
--> 935         offset = self._particle_blocks.index(blockname)
    936         _type = self._particle_types[offset]
    937 

ValueError: 'metal' is not in list
```

The reason is this snippet of code
https://github.com/pynbody/pynbody/blob/aa4bd2a0e9fb411d68c246daac7a2752e0020d2f/pynbody/snapshot/ramses.py#L1070-L1077
assumes that `tform` and `metal` are always present rather than checking that they are available to be loaded, triggering a `ValueError` which is not caught here
https://github.com/pynbody/pynbody/blob/aa4bd2a0e9fb411d68c246daac7a2752e0020d2f/pynbody/snapshot/__init__.py#L364-L369
and crashes the code before any further attempts at deriving the array here
https://github.com/pynbody/pynbody/blob/aa4bd2a0e9fb411d68c246daac7a2752e0020d2f/pynbody/snapshot/__init__.py#L356-L358

I think the cleanest fix is to encapsulate dealing with `metal` and `tform` inside this test 
https://github.com/pynbody/pynbody/blob/aa4bd2a0e9fb411d68c246daac7a2752e0020d2f/pynbody/snapshot/ramses.py#L1083-L1086
to ensure that the code only gets triggered if these fields are actually available. I can also adds a trivial test that copies one of the existing Ramses snapshot, but changing its header to random keys to check for this behaviour.

Any opinions?

Martin

Martin-Rey:
This should fix #689 and I added a test that used to fail with the previous behaviour but is now passing.

The original pynbody behaviour is probably not immediately problematic for most users, because it would fail with a somewhat useful `ValueError: 'metal' is not in list` when trying to access `tform` or `metal` that are not present on disc. 

However, it triggers a `ValueError` due a coding oversight that didn't check that the fields were available to load, rather than the wanted `KeyError: No array metal on disc`.

This is problematic when trying to define your own `metal` array as a derived property (in my case, I have Fe, O, Al, ... on disc, which I aggregate into `metal`). The derived array would then also fail, because pynbody then fails to recognise that loading from disc didn't work, does not attempt to derive the array and simply crash with the `ValueError`.

Hope that's clear.
Martin


Martin-Rey:
pre-commit.ci autofix

apontzen:
@Martin-Rey could you resolve the merge conflicts here?

PR #696, opened 2022-12-15
Fixes and test the conversion from Ramses to Tipsy

Martin-Rey:
Hi, 

This PR implements a few changes to make the converter from Ramses to Tipsy more robust and adds a test that verifies that the newly converted snapshot has units and numbers matching the original. Features are:

- The converter can now detect whether the simulation is DMO, or contains gas, and deal with the conversion accordingly
- It now converts the potential in the right units, which would mess up Ramses runs that explicitly store `phi`
- I updated the Tipsy writing to ensure that `ntot` in the Tipsy header reflects the number of gas, DM and star particles that are going to be written to disc, triggering a warning when doing so. I am a little unsure about this, but the `write()` does not seem setup to deal with other families (like ramses tracers, sinks etc) anyway, and having a `ntot` different from the number of particle actually written in the file would mess up later analysis tools that rely on this information to figure out the length of particle blocks.

The test is extensive on a DMO and Hydro runs with the old Ramses format. There are currently no cosmological output with the new format, but I have verified on my data that the test also passes.

Let me know your thoughts


Martin-Rey:
pre-commit.ci autofix

apontzen:
I suggest merging master into here to fix the failing tests

apontzen:
Thanks!

Issue #691, opened 2022-11-18
Particle family mappings of AHF haloes


Martin-Rey:
Hi @apontzen and @cphyc 

I am having some interesting behaviours comparing my simulated data in a Ramses and Tipsy formats, both with the same AHF catalogue. When loading the snapshot as Tipsy, the particle families of the halo are well setup and match the AHF record
```
In [27]: type(s)
Out[27]: pynbody.snapshot.tipsy.TipsySnap

In [23]: s.halos()[1].properties['npart']
Out[23]: 7332330

In [24]: s.halos()[1]
Out[24]: <SimSnap "./output_00113_tipsy/output_00113_fullbox.tipsy:halo_1" len=7332330>

In [19]: s.halos()[1].properties['n_gas']
Out[19]: 3608832

In [25]: s.halos()[1].g
Out[25]: <SimSnap "./output_00113_tipsy/output_00113_fullbox.tipsy:halo_1::gas" len=3608832>

In [18]: s.halos()[1].properties['n_star']
Out[18]: 3955

In [26]: s.halos()[1].st
Out[26]: <SimSnap "./output_00113_tipsy/output_00113_fullbox.tipsy:halo_1::star" len=3955>

In [39]: s.halos()[1].dm
Out[39]: <SimSnap "./output_00113_tipsy/output_00113_fullbox.tipsy:halo_1::dm" len=3719543>
```

However, when I repeat the same operation sloading my simulation data as a Ramses snapshot, the stars are wrongly mapped onto the DM family, yielding 0 stars associated with the halo and too many DM particles (which I guess is inferred as the difference between `npart`, `n_stars` and `n_gas`).

```

In [29]: type(s)
Out[29]: pynbody.snapshot.ramses.RamsesSnap

In [30]: s.halos()[1].properties['n_star']
Out[30]: 3955

In [31]: s.halos()[1].st
Out[31]: <SimSnap "./output_00113:halo_1::star" len=0>

In [32]: s.halos()[1].properties['n_gas']
Out[32]: 3608832

In [34]: s.halos()[1].g
Out[34]: <SimSnap "./output_00113:halo_1::gas" len=3608832>

In [35]: s.halos()[1].properties['npart']
Out[35]: 7332330

In [36]: s.halos()[1]
Out[36]: <SimSnap "./output_00113:halo_1" len=7332330>

In [37]: s.halos()[1].dm
Out[37]: <SimSnap "./output_00113:halo_1::dm" len=3723498>

```

[...]

PR #697, opened 2022-12-20
Fix family mapping for AHF halos

Martin-Rey:
Hi @apontzen and @cphyc,

I finally got to the bottom of #691, which I have now corrected in this PR. The summary is: AHF is typically run on snapshots that have DM+star+gas and nothing else. New Ramses snapshots (and maybe others) have extra families (BHs, tracers, ...), adding a lot of additional particles that AHF is unaware of. When reading the AHF membership IDs on disc, one then need to remap them to their correct family slices, taking into the extra particles into account.  

They are two open points for me in this
1. I am not sure whether the explicit type check on `RamsesSnap` and `NchildaSnap` are necessary. These seem to have been introduced to deal with the fact that sometimes, gas needs to be the first family, whereas in some others, it's the last. I feel like `self.base._get_family_slice` should deal with this naturally without having to know the snapshot type. I'm happy to try removing these type checks, but I don't think they are tested explicitly, and don't want to break others' code without knowing.
2. I constructed a test based on an EDGE2 snapshot, to help me isolate the issue and convince myself it is fixed. The testdata is currently missing and failing the PR, but is also a fairly beefy production output.  I would prefer to have a more portable test snapshot, but would need an output with the new Ramses format and particle tracers that is small and complete to be able to run AHF on it (unlike the test snapshot which is missing CPU files). What do you think is the best way forward for this?

Hope that helps,
Martin


apontzen:
I suggest merging master into here to fix the failing tests

Martin-Rey:
I would be happy to do this, but we also need to update the testdata to add an additional snapshot to fix the tests. It's unclear to me what the best way forward for this is (Point 2 above).

apontzen:
I suggest merging anyway, so that we can see that it's only those tests missing data which are failing.

What size are we talking about? Is it truly impossible to hack it to work with missing CPUs?

Martin-Rey:
That should be the case now, let m play around with the formats and outputs to see how I can compact them.

Martin-Rey:
Ok I have managed to reduce the target file to 1.2G and to produce an AHF output for it. I need a decent amount of particles to produce a halo catalogue that makes sense and that be can interpreted and tested.

I haven't managed to convert and/or run AHF on the existing snapshots, as the only output with the new format is non-cosmological and returns an empty halo catalogue.

Let me know if that works and I'll send you a gzip.

Martin-Rey:
Hi Andrew, as far as I can tell, the tests are now failing solely because of the lack of additional test data.

Let me know if the above addition sounds reasonable to you.

apontzen:
Please do point me to the new dataset you created for testing. I am concerned at the resulting big increase in the testdata tar size, but I can’t really think of an alternative.

apontzen:
I added the file but it looks like the path isn't yet correct in the test suite

Martin-Rey:
pre-commit.ci autofix

apontzen:
Regarding the outstanding matter of general vs format-specific logic: I agree on both fronts that it would be better if generalised, but I prefer not to make potentially breaking changes unless we have a lot of time to test the potential implications.

Everything else looks fine to me so I am happy to merge this if you agree it's ready?

Martin-Rey:
Sounds good, please go ahead and merge

PR #701, opened 2023-01-23
Update `pypi.yaml` action dependencies and add dependabot to automatically keep them up to date

jobovy:
I noticed that the action dependencies in `pypi.yaml` were out of date, which meant, for example, that you weren't building a Python 3.11 wheel (because the `cibuildwheel` action version was old). This updates the dependencies to their latest versions and additionally adds a `dependabot.yml` file that should keep them automatically up to date (I put a monthly schedule to check such that this doesn't cause too many PRs).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    interval: "monthly"
--- a/.github/workflows/pypi.yaml
+++ b/.github/workflows/pypi.yaml
@@ -3,6 +3,11 @@ name: Push to PyPI
 on:
   push:
     tags: v*
+  pull_request:
+    branches:
+      - master
+    paths:
+      - '.github/workflows/pypi.yaml'
 
   workflow_dispatch:
 
@@ -43,11 +48,13 @@ jobs:
           brew install gcc@10
 
       - name: Determine version tag
+        if: github.event_name != 'pull_request'
         run: |
           echo "SETUP_VERSION=`python setup.py --version`" >> $GITHUB_ENV
           echo "VERSION_TAG=`git describe --tags | cut -c 2-`" >> $GITHUB_ENV
 
       - name: Verify version naming is consistent
+        if: github.event_name != 'pull_request'
         run: |
           if [ "${{ env.VERSION_TAG }}" == "${{ env.SETUP_VERSION }}" ]; then
               echo Git tag and python setup.py versions match: ${{ env.VERSION_TAG }}
@@ -68,6 +75,7 @@ jobs:
   build_sdist:
     name: Build source distribution
     runs-on: ubuntu-latest
+    if: github.event_name != 'pull_request'
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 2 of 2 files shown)


jobovy:
pre-commit.ci autofix

jobovy:
I don't think the tests run here actually test the changes in this PR, because the `pypi.yaml` workflow isn't run on PRs. I think the changes should be fine, but it might be prudent to run the wheel building on PRs to make sure they succeed and just not upload them to PyPI. I can add that to this PR if you want.

apontzen:
Thanks Jo. I have a manual trigger (`workflow_dispatch`) on `pypi.yaml` for precisely this reason, but it doesn't seem to let me run it on your branch for some reason. Do you happen to have any ideas why? (I don't want it always to build for every PR because it takes a very large number of core hours and I feel bad about the CO2...)

jobovy:
pre-commit.ci autofix

jobovy:
Okay, that makes sense, the wheels do seem to take a long time to build. To help with this, I've done the following:
* Run the `pypi.yaml` workflow on PRs, but only those that change `pypi.yaml`; those should be rare, but at least make sure that changes to `pypi.yaml` don't break the wheel-building infrastructure
* I've dropped 32 bit Linux wheels and the PyPy wheels, which I think you were only building because they are included by default by `cibuildwheel`? I don't think many people use those. Instead, I've switched to a similar setup as, e.g., `numpy` and `scipy` use, building 64 bit Linux and Mac wheels for CPython >= 3.8. Also split each into its own job so they run more efficiently. This seems to significantly speed things up. By adding to the strategy matrix, you can always add these back in if people ask for them.

You might want to consider adding Mac arm64 wheels, which should be as easy as adding the line
```
- [macos-latest, macosx, arm64]
```
in the `buildplat` part of the `matrix` (but I haven't tested this).

Also cleaned up the history a bit. However, you might still just want to squash and merge. I think it's ready to be merged now.

Also, if you're concerned about runtime, you might want to cache the test-data downloads, because those seem to take a long time and should be easy to cache using `actions/cache`.

apontzen:
Thank you Jo, I agree this is a much improved workflow.

I would love it to build for arm64 on mac, but I believe the cross-compilation is not possible if one also uses OpenMP. This might be a historical thing, perhaps it's sorted out now. I'll take a look.

I agree about caching test data downloads -- for some reason I couldn't get this to work when I tried in the past, I'm sure due to user error...

Shall I merge your changes for now and look at the other matters separately?

jobovy:
Yes, please merge this. I don't have time to look into the other suggestions I made (and as you say, perhaps the arm64 build is more complicated) and they would be best done in a separate PR anyway. Thanks!

FYI: I noticed that you remove the code in `setup.puy` that tried to support `clang`. In `galpy`, I found a pretty robust solution for turning off the `-fopenmp` flag by using a builder extension that checks for each flag whether it's supported:
https://github.com/jobovy/galpy/blob/dc099f2b97312cb2e97b84118aee45d2db69f122/setup.py#L248
The advantage of this method is that you're sure that you are checking the compiler that's actually used to compile the code. Maybe something to keep in mind if your user base would like `pynbody` to run on ARM Macs (I don't need it).

apontzen:
Thanks, all very helpful!

PR #702, opened 2023-01-26
Replace `xdrlib` use for reading Nchilada with direct implementation

### LINKED ISSUE #700: `xdrlib`
**Describe the bug**

`xdrlib` is deprecated and slated for removal in Python 3.13 (see [PEP 594](https://peps.python.org/pep-0594/#xdrlib); less than two years from now). Used in
https://github.com/pynbody/pynbody/blob/2f6647b93384583c7275ef190d814a8d3ec7826d/pynbody/snapshot/nchilada.py#L14

**To Reproduce**

```
from pynbody.snapshort import nchilada
```
raises a DeprecationWarning.

No DeprecationWarning

 - Python 3.11
 - pynbody: 1.3.0

**Additional context**

Issue popped up in `galpy`'s CI (see, e.g., [here](https://github.com/jobovy/galpy/actions/runs/3987639158/jobs/6837735334#step:20:65)), where I now turn all DeprecationWarnings into errors to guard against future issues. I'm sure I can work around this....

jobovy:
Fixes #700.

I looked at the relevant `xdrlib` code and it seemed very easy to replace it with a direct implementation, so I've quickly done that here (also my first ever use of the walrus operator, which you can use because you require Python >= 3.8!). I haven't tested this locally, because I don't actually know how to use any of the files involved, but the test suite runs successfully and it looks like it should test this code as part of the Nchilada test? Anyway, you or somebody who uses these files might want to check locally yourself to make sure that this works.

### DIFF
--- a/pynbody/snapshot/nchilada.py
+++ b/pynbody/snapshot/nchilada.py
@@ -10,8 +10,8 @@ automatically via pynbody.load.
 """
 
 import os
+import struct
 import warnings
-import xdrlib
 import xml.dom.minidom
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


apontzen:
Thanks Jo, looks great. I confirm this does get tested. Would you mind refactoring this into a separate function, though, just to separate out the low-level IO from the higher-level reading logic?

apontzen:
Yes, fair enough. Thanks!

PR #703, opened 2023-01-27
Investigate whether we can build macos arm64 wheels

apontzen:


apontzen:
pre-commit.ci autofix

Issue #705, opened 2023-02-06
pynbody-meta dtype information writing is incompatible with reading


trquinn:
Using the template below will help us reproduce your problem, which is an essential first step in helping or improving the code. Feel free to customise the template if you are used to raising bug reports, but be sure that you give us enough information to start investigating.

**Describe the bug**
The field for "dtype" in the pynbody-meta file for Tipsy arrays is being written as (e.g.):
`dtype:` <class 'numpy.int32'>`
This can't be parsed when the array file is loaded.  This becomes a problem when writing/reading binary integer arrays.

**To Reproduce**
To see a problem requires a tipsy snapshot with a non-standard integer array.  A common example is `*.amiga.grp`.   Such a snapshot can be read in if the tipsy array files are ASCII.  If the snapshot is read in with `sim = pynbody.load()`, and  `sim['amiga.grp']`,  then writen out with `sim.write(filename='new_file',fmt = pynbody.tipsy.TipsySnap,binary_aux_arrays=True)`, reloading from the new file will produce garbage values for the `amiga.grp` property, and its dtype will be 'float32'.

**Expected behaviour or result**
I expect the new snapshot created as above to have the same values for amiga.grp, and amiga.grp to be integer valued.

**Setup (please complete the following information):**
 - OS CentOS Linux release 7.9.2009 (Core)
 - Python version Python 3.7.0 
 - Numpy version numpy-1.16.1-py3.7
 - Pynbody version 1.2.3

PR #706, opened 2023-02-06
TipsySnap._write_array_metafile(): write dtype correctly.

trquinn:
This also completes documentation to the _get_array_metafile() method.

Addresses issue #705


trquinn:
This is always an issue on the big compute resources: they are very conservative in upgrading packages.  Such that I can't build the latest version of pynbody on them.

apontzen:
pre-commit.ci autofix

PR #708, opened 2023-03-03
Cache testdata in GitHub Actions build-test workflow

jobovy:
As I mentioned in https://github.com/pynbody/pynbody/pull/701#issuecomment-1401097177, caching the testdata should speed up the build-test workflow a bit. Because I just had to implement a similar directory caching in another project (and learned how easy it really is), I thought I might as well add it here too. So this PR caches the `testdata` download step. As people have pointed out elsewhere, this really just replaces downloading from the web with downloading from GitHub's servers, but it does seem to be substantially faster (~45s vs. 2.5 minutes).

I believe the testdata rarely changes? So the cache as written never expires. If you have to update it, either change the version in the cache key or just delete the cache on the Actions tab.

I'll push another commit getting rid of your source `h5py` installation for Python 3.11 to illustrate how the cache is used; ``h5py`` now has Python 3.11 wheels, so the source compilation is no longer necessary.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -44,11 +44,6 @@ jobs:
       run: |
         python -m pip install --upgrade pip setuptools wheel
         python -m pip install numpy${{ matrix.numpy-version }}
-    - name: Manually build h5py for python 3.11
-      if: ${{ matrix.python-version == '3.11' }}
-      run: |
-        sudo apt-get install libhdf5-dev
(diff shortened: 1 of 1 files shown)


jobovy:
Okay, the second run shows that the cache is being used, e.g., https://github.com/jobovy/pynbody/actions/runs/4320370356/jobs/7540549341

The `pre-commit` checks pass locally for me, but to fix the failure here you will have to merge #704.

apontzen:
Thank you!

PR #715, opened 2023-04-13
Clarify intent of use_family vs only_family when matching halos

apontzen:
Add test that matching on families actually works


apontzen:
pre-commit.ci autofix

apontzen:
Approved offline by @mtremmel

PR #731, opened 2023-09-08
Switch to using numpy.get_include() to get numpy header files

jobovy:
``numpy.distutils`` is being removed in Python 3.12, so this PR replaces its use in ``pynbody``'s ``setup.py`` with ``numpy.get_include``, which has identical functionality and has been in ``numpy`` since, I believe, version 1.6, so a long time ago.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ import sys
 import tempfile
 from os import path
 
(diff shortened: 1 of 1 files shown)


apontzen:
Thanks!

PR #732, opened 2023-09-08
Test build with python 3.12

apontzen:


apontzen:
Note that at the moment too much is broken upstream to get pynbody working with 3.12.

At the heart of the problem is that numpy is changing its approach to building:

https://github.com/numpy/numpy/issues/23808

While this can be resolved by pointing pynbody at numpy 1.26.0rc1, other upstream packages including h5py and matplotlib do not yet install successfully.

Python 3.12 compatibility should be revisited once these packages support it.

jobovy:
Just stumbled on this while looking for the v1.3.2 wheels (are you going to push those to PyPI?). I'm actually able to build and run `pynbody` under Python 3.12 using Ubuntu in `galpy`'s CI, relevant setup is here:

https://github.com/jobovy/galpy/blob/2b87b18abc8542b5b96126008848b0e9801d1b3d/.github/workflows/build.yml#L245

For macs, the issue is that I can't get `h5py` to compile, so there I probably have to wait until there are `h5py` wheels. But you should be able to test in general whether `pynbody` runs under Python 3.12.

apontzen:
Thanks Jo. I'll probably just wait until h5py sorts it out, otherwise it'll be a painful two-step process. It looks like wheels for h5py on python 3.12 are coming any day now.

jobovy:
I see now, that’s great, last time it took months to get `h5py` wheels.

PR #740, opened 2023-10-24
Fix time conversion

cphyc:
The current state of RAMSES is quite a mess. Namely, times that are written on disk may be one of the following:
- proper physical times, with 0 at the beginning of the simulation. This is the default for non-cosmological simulations.
- conformal times. This is the default for cosmological simulations.
- proper physical times, in units of 1/H0. This is e.g. the default for RT simulations, such as EDGE2.

As usual, there isn't much in the data to automatically detect case 2 from case 3...

In this PR, I cleaned up the code so that one can write in `.pynbodyrc` the following
```
[ramses]
proper_time: true
```

This will force the use of option 3 over option 2. I also ensured that all time conversions were consistent throughout and cleaned up the mess that the RAMSES frontend had become.


pynbody/snapshot/ramses.py
```diff
@@ -44,7 +44,6 @@
     try:
         import multiprocessing
 
-        import posix_ipc
```
apontzen, line 47: This should probably have a noqa flag on it - while posix_ipc is not technically needed in this scope, the ability to import it is being tested here

apontzen:
Thanks for all this work -- what a huge pain this turns out to be!

Regarding detecting what's going on, if I recall correctly the conformal tforms are distinguished by being negative. So is it possible to use the existing "is cosmological" logic, then look for negative tforms:

1) If cosmological + negative tforms, apply map from conformal to real times

2) If cosmological + positive tforms, apply H0 unit conversion

3) Otherwise, leave alone

Of course, a configuration option and/or a kwarg to the initialiser to override all this guesswork is also advisable, but would this be ok for most people?

cphyc:
> Thanks for all this work -- what a huge pain this turns out to be!
> 
> Regarding detecting what's going on, if I recall correctly the conformal tforms are distinguished by being negative. So is it possible to use the existing "is cosmological" logic, then look for negative tforms:
> 
>     1. If cosmological + negative tforms, apply map from conformal to real times
> 
>     2. If cosmological + positive tforms, apply H0 unit conversion
> 
>     3. Otherwise, leave alone
> 
> 
> Of course, a configuration option and/or a kwarg to the initialiser to override all this guesswork is also advisable, but would this be ok for most people?

This was my initial guess, unfortunately, for cosmological simulations, the times written on disk are always negative. Indeed, when written in proper units, the zero point is *at z=0*. That means that times are actually time-to-redshift zero rather than time-since-big-bang. For conformal units, we also have negative values, with 0 corresponding to particles formed at z=0.

They're unfortunately no robust way to distinguish these times because there might or not be stars formed close to the beginning of the simulation where the difference is the largest. Instead of implementing a somewhat clunky guess, I added a way of building the RAMSES snap object and forcing it to assume times are in proper units.

apontzen:
Thanks, I see! Did you say there is a link between using the RT module and the distinction between physical / conformal formation times? Is that worth using to provide a default, potentially while emitting a warning in case people's conventions differ?

If this is also no good as even a rough predictor then I agree all we can do is provide the option. Which variant is likely to be most common amongst simulations people are running today? I reckon that most common variant should be the default, and a warning should be issued unless people provide the flag.

cphyc:
The defaults are:
- non cosmo simulations: proper time;
- cosmo simulations: conformal time;
- cosmo simulations with RT: proper time (but offset such that t=0 for z=0).

At the moment, the first two options are activated by default. However, there's no logic implemented in pynbody to detect RT, so I'm a bit reluctant to do so. And also, if for some reason the `rt` files are missing, we'd guess the wrong thing.

apontzen:
Is there anything wrong with looking for an RT file, if one is found defaulting to proper time, if one isn't found defaulting to conformal time? In both cases, a warning could be issued unless the user has explicitly passed in a flag.

I feel overall this is the 'least likely to result in catastrophic errors' option...

cphyc:
@apontzen I've implemented the logic of your comment. I've added a test to verify that RT simulations are correctly parsed as proper times, despite being cosmological.

The test is however failing due to another unrelated bug (#742) which I haven't been able to fix just yet.

apontzen:
I opened a pull request to your pull request ;-) https://github.com/cphyc/pynbody/pull/3 

Hopefully this fixes the immediate issue, and I think #742 can be considered separately? (See my comment on that issue)

PR #745, opened 2023-11-02
Fix warnings

cphyc:
There used to be a very large number of warnings being thrown in the test suite.

Most of them are related to unclosed file descriptors.

In this PR, I either catch or I fixed the closing of the files.


apontzen:
I've merged this into #749 and fixed it there. Thanks for all the work on this.

Issue #746, opened 2023-11-02
Tests fail if ran in random order


cphyc:
**Describe the bug**
When running the test suite, some tests fail when run in a given order.

**To Reproduce**

Install pytest-random-order (`pip install pytest-random-order`) then run the tests:
```bash
# This succeeds
pytest --random-order --random-order-seed=509902 tests/nchilada_test.py -v
# This fails
pytest --random-order --random-order-seed=1234 tests/nchilada_test.py -v
```

Output:
```
================================================================================================ test session starts ================================================================================================
platform linux -- Python 3.12.0, pytest-7.4.3, pluggy-1.3.0 -- REDACTED/py312/bin/python3.12
cachedir: .pytest_cache
Using --random-order-bucket=module
Using --random-order-seed=1234

rootdir: /home/cphyc/Documents/prog/pynbody
configfile: pyproject.toml
plugins: random-order-1.1.0
collected 5 items                                                                                                                                                                                                   

tests/nchilada_test.py::test_get PASSED                                                                                                                                                                       [ 20%]
tests/nchilada_test.py::test_HI_not_present_for_non_gas_particles PASSED                                                                                                                                      [ 40%]
tests/nchilada_test.py::test_partial_loading PASSED                                                                                                                                                           [ 60%]
tests/nchilada_test.py::test_array_completion_unit_sanity PASSED                                                                                                                                              [ 80%]
tests/nchilada_test.py::test_get_gas PASSED                                                                                                                                                                   [100%]

================================================================================================= 5 passed in 0.88s =================================================================================================


================================================================================================ test session starts ================================================================================================
platform linux -- Python 3.12.0, pytest-7.4.3, pluggy-1.3.0 -- /home/cphyc/miniconda3/envs/py312/bin/python3.12
cachedir: .pytest_cache
Using --random-order-bucket=module
Using --random-order-seed=509902

rootdir: /home/cphyc/Documents/prog/pynbody
configfile: pyproject.toml
plugins: random-order-1.1.0
collected 5 items                                                                                                                                                                                                   

tests/nchilada_test.py::test_array_completion_unit_sanity FAILED                                                                                                                                              [ 20%]
tests/nchilada_test.py::test_HI_not_present_for_non_gas_particles PASSED                                                                                                                                      [ 40%]
tests/nchilada_test.py::test_partial_loading PASSED                                                                                                                                                           [ 60%]
tests/nchilada_test.py::test_get PASSED                                                                                                                                                                       [ 80%]
[...]

PR #747, opened 2023-11-10
Fix issue #746

apontzen:
Modernize nchilada tests and in particular don't pass information between tests!


apontzen:
Merged into #749

Issue #726, opened 2023-06-30
Trying to run the AHF halo finder throws a configparser InterpolationSyntaxError


bwkeller:
I'm not sure if this is because of a change in Python3's configparser library (I recently upgraded on my machine) or a change that's crept into to pynbody, but recently trying to run AHF automatically by calling `sim.halos()` crashes on line 734 in ahf.py, with the following traceback:

```
File ~/.local/lib/python3.11/site-packages/pynbody/halo/ahf.py:734, in AHFCatalogue._run_ahf(self, sim)
    730 lgmax = np.min([int(2 ** np.floor(np.log2(
    731     1.0 / np.min(sim['eps'])))), 131072])
    732 # hardcoded maximum 131072 might not be necessary
--> 734 print(config_parser.get('AHFCatalogue', 'Config', vars={
    735     'filename': sim._filename,
    736     'typecode': typecode,
    737     'gridmax': lgmax
    738 }), file=f)
    740 print(config_parser.get('AHFCatalogue', 'ConfigTipsy', vars={
    741     'omega0': sim.properties['omegaM0'],
    742     'lambda0': sim.properties['omegaL0'],
   (...)
    746     'eunit': 0.03  # surely this can't be right?
    747 }), file=f)
    749 f.close()

File /usr/lib/python3.11/configparser.py:815, in RawConfigParser.get(self, section, option, raw, vars, fallback)
    813     return value
    814 else:
--> 815     return self._interpolation.before_get(self, section, option, value,
    816                                           d)

File /usr/lib/python3.11/configparser.py:396, in BasicInterpolation.before_get(self, parser, section, option, value, defaults)
    394 def before_get(self, parser, section, option, value, defaults):
    395     L = []
--> 396     self._interpolate_some(parser, option, L, value, section, defaults, 1)
    397     return ''.join(L)

File /usr/lib/python3.11/configparser.py:428, in BasicInterpolation._interpolate_some(self, parser, option, accum, rest, section, map, depth)
    426 m = self._KEYCRE.match(rest)
    427 if m is None:
--> 428     raise InterpolationSyntaxError(option, section,
    429         "bad interpolation variable reference %r" % rest)
    430 var = parser.optionxform(m.group(1))
    431 rest = rest[m.end():]

[...]

Issue #434, opened 2018-02-01
invoking AHF not working


TobiBu:
I pulled the latest version of pynbody and now it does not invoke AHF anymore.

I get the following error message:

`RuntimeError: Cannot guess the HOP catalogue filename for /disk2/buck/PhD/HYDRA/7.66e11/7.66e11.00632`

reversing to commit 04e94c72ec4fc61c66bcc9099ccb8c7a5c48bf5a from September before work on the HOP catalogue was done fixes the problem.

apontzen:
Thanks, could you check if this is fixed by branch `issue-434`?

TobiBu:
Yes, this branch fixes the bug.

TobiBu:
I guess the bug is not fully fixed yet. It invokes AHF now correctly however, when trying to asses the luminosity of stellar particles I get the following error:

`---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-19-579c6d271d22> in <module>()
----> 1 h[1].s['v_mag']

/home/Buck/.local/lib/python2.7/site-packages/pynbody-0.42-py2.7-linux-x86_64.egg/pynbody/snapshot/__init__.pyc in __getitem__(self, i)
    261 
    262         if isinstance(i, str):
--> 263             return self._get_array_with_lazy_actions(i)
    264         elif isinstance(i, slice):
    265             return SubSnap(self, i)

/home/Buck/.local/lib/python2.7/site-packages/pynbody-0.42-py2.7-linux-x86_64.egg/pynbody/snapshot/__init__.pyc in _get_array_with_lazy_actions(self, name)
    354                     self.__load_if_required(name)
    355                 if not self.lazy_derive_off:
--> 356                     self.__derive_if_required(name)
    357 
    358             return self._get_array(name)

/home/Buck/.local/lib/python2.7/site-packages/pynbody-0.42-py2.7-linux-x86_64.egg/pynbody/snapshot/__init__.pyc in __derive_if_required(self, name)
    369     def __derive_if_required(self, name):
    370         if name not in self.keys():
--> 371             self._derive_array(name)
    372 
    373     def __resolve_obscuring_family_array(self, name):

/home/Buck/.local/lib/python2.7/site-packages/pynbody-0.42-py2.7-linux-x86_64.egg/pynbody/snapshot/__init__.pyc in _derive_array(self, array_name, fam)
   1939     def _derive_array(self, array_name, fam=None):
[...]

PR #763, opened 2024-01-19
Modernize and streamline HaloCatalogue class hierarchy

apontzen:
Work on tangos has revealed there are deep-seated issues with the way the halo catalogue system works, with a jumble of approaches and different interfaces.

I am using this opportunity to largely re-write the halo catalogue base class and adjust/simplify the child classes. This clearly comes with risks attached but a modernisation is necessary for tangos and hopefully good for everyone in the long run.

I will also use this opportunity to remove the ability to run halo finders, which was always rather scattergun and goes beyond the boundaries of pynbody's functionality elsewhere. (See #434, #726). 

Implementations to complete:

- [x] grp / amigagrp 
- [x] Adaptahop
- [x] AHF
- [x] hop
- [x] rockstar
- [x] subfind
- [x] subfindhdf

Refinements to make:

- [x] iord to fpos mapping is probably better implemented as a clear class hierarchy, rather than pretending to always be an array. It needs better clarity e.g. over retaining ordering, and possibly profiling to see whether it is more efficient to use `numpy.searchsorted` [completed except for profiling which can be tackled if needed another time]
- [x] with clearer iord to fpos mapping, do an audit of where this is used in the codebase, and when it is applied to the whole catalogue vs to single halos
- [x] The `sub` mechanism implemented in subfindhdf should be clarified and extended to other relevant cases
- [x] tracking subclasses of HaloCatalogue should be automated, removing the need for a manually maintained list
- [x] users should be able to pass their own class priority list to `SimSnap.halos`, overriding the config-determined ordering
- [x] method naming and typing of the protected members of `HaloCatalogue` should be reviewed

Question: can we remove the `halo.legacy` module (@mtremmel ?)


apontzen:
Additional question for @mtremmel 

It seems to me like when AHF halos are loaded currently, their ID value (if specified in the AHF halos file) is essentially ignored. If I am understanding correctly, this has led to a lot of heartache with tangos.

One possibility is that we actually start using the AHF-provided ID value in pynbody, rather than assuming just that the IDs are 1...N. Would this be a welcome change? 

I am not actually clear why/when AHF generates IDs that are not sequential. It's obviously a breaking change and it would perhaps be necessary to introduce it as an option to start with, and over a couple of years switch it to being the default option.

Thoughts very welcome (also @EthTay and @cphyc who I know have played with AHF considerably)

apontzen:
Thanks for the explanation @mtremmel .

The problem we have at the moment is the different halo loaders in pynbody behave very differently, but if we expose more of this to the user and make things work similarly under the hood we should be ok.

I'm working on it, in a way that will break as little as possible ;-)

apontzen:
There is a linked problem which is... why on earth do we start numbering things at 1, not 0? I am guessing this is some throwback to alyson's IDL tools. But even in the test data file (g15784), AHF seems implicitly itself to number things starting at zero.

mtremmel:
Yeah... I think this is just tradition to be honest...

PR #767, opened 2024-02-10
Update ahf.py

EthTay:
substructure file failing to be read correctly due to an update in a way the file is being read, this would throw an error and then default back to reading the halo file,

this update correctly adresses the way the file is now being read and stored


apontzen:
Is it urgent to fix this? Completely re-implemented halo readers are on their way (#763), but the timescale is a bit unclear.

The failing unit tests seem to be related to upstream dependencies in a variety of ways, some of which I understand and others don't. I'm currently investigating in #769 .

apontzen:
Superseded by #763

PR #782, opened 2024-03-15
Changed halo shape function

Morkney:
I've amended the previous "halo_shape" function into a new "shape" function.

This new version is generalised with respect to particle type (i.e. not limited to dark matter), and can handle both 3D and 2D shape calculations. I think it is more useful this way.

There are also some other minor improvements and simplifications.

### DIFF
--- a/pynbody/analysis/halo.py
+++ b/pynbody/analysis/halo.py
@@ -403,9 +403,9 @@ def shape(sim, nbins=100, rmin=None, rmax=None, bins='equal', ndim=3, max_iterat
   assert rmin >= 0
   assert rmax > rmin
   assert nbins > 0
-  if ndim==3:
+  if ndim==2:
     assert np.sum((sim['rxy'] >= rmin) & (sim['rxy'] < rmax)) > nbins*2
-  elif ndim==2:
+  elif ndim==3:
     assert np.sum((sim['r'] >= rmin) & (sim['r'] < rmax)) > nbins*2
   if bins not in ['equal', 'log', 'lin']: bins = 'equal'
 
--- a/pynbody/analysis/halo.py
+++ b/pynbody/analysis/halo.py
@@ -338,151 +338,197 @@ def center(sim, mode=None, retcen=False, vel=True, cen_size="1 kpc", move_all=Tr
 
     return tx
 
-def halo_shape(sim, N=100, rin=None, rout=None, bins='equal'):
-    """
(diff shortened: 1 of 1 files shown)


apontzen:
Thanks Matt. Could you add a test that checks the results, e.g. on the `g15784.lr.01024` galaxy in the test data?

Morkney:
> Thanks Matt. Could you add a test that checks the results, e.g. on the `g15784.lr.01024` galaxy in the test data?

I've added a test for 2D & 3D. Let me know if they're not suitable!
There's some failures on the "pre-commit.ci - pr" checks, but I don't really understand these.
Best, Matt

apontzen:
pre-commit.ci autofix

apontzen:
Sorry for the delay in getting this merged. I'm a bit confused actually about the changes in the return arrays. The doc says

```
 *axis lengths*: The axis lengths of each homoeoidal shell in
   order a>b>c with units of ``sim['pos']``.
```
However in the code the array being returned is called `axial_ratios` and accordingly is of size Nx2 not Nx3. 

What actually is being returned and how should the user interpret it?


On a separate note, I am concerned that changing the name and interface may unnecessarily break things for other users (paging @bvogel4 for an opinion on that). One option would be to introduce a deprecated wrapper function that presents the same interface to existing users.

Morkney:
Hello,

I changed the name only because "**halo**_shape" seemed redundant if the user would also be using the function for, e.g., stars. I don't have any real preference, though. If it's simpler to leave the name as it is, then that is fine with me!

You are right @apontzen to point out that "axial_ratios" is not descriptive, it should be "axis_lengths". I think axial_ratios is the legacy name of the array and I forgot to change it. The array should be of size (nbins,ndim) where nbins is the number of radial shells, and ndim is the number of spatial dimensions which can be either 2 or 3.
The user should interpret the output as an array with one column for each of the axial lengths (either a,b or a,b,c) and with one row for each radial shell, corresponding to the "rbins" output.

I think it is better to format the output in this way, and then allow the user to calculate the axial ratios as they wish. I don't mind if other users would prefer the format to match that of the original function, though that format may be more confusing for anybody who wants to calculate the 2D projected shapes. A wrapper would solve this, I suppose.

apontzen:
Sounds like a good idea but we should go one step at a time. First I need to merge this in, but that requires working on resolving conflicts and on conforming the docs. Sorry this isn't at the top of my list right now, but hopefully I will get to it in the next few days.

Issue #762, opened 2024-01-19
RuntimeErrors upon interpreter shutdown due to garbage collection


pavyamsiri:
**Describe the bug**
There are various runtime errors when a snapshot is being garbage collected during shutdown.

**To Reproduce**
```python
import sys
import pynbody

def main():
    snapshot_path = sys.argv[1]
    snapshot = pynbody.load(snapshot_path)

if __name__ == "__main__":
    main()
```

I get this when opening certain Ramses snapshots, for example in the pynbody testdata, opening `ramses_new_format_partial_output_00001`:
**Expected behaviour or result**
The expected behaviour is no warnings or errors when simply loading a snapshot and exiting. Instead I get
```python
Exception ignored in: <function _get_auto_closing_shared_memory_buffer.<locals>.clearup_callback at 0x7f57579696c0>
Traceback (most recent call last):
  File "/home/pavy/micromamba/envs/phd/lib/python3.12/site-packages/pynbody/array/shared.py", line 169, in clearup_callback
    t.start()
  File "/home/pavy/micromamba/envs/phd/lib/python3.12/threading.py", line 992, in start
    _start_new_thread(self._bootstrap, ())
RuntimeError: can't create new thread at interpreter shutdown
```
which appears multiple times (every time a shared memory array is getting garbage collected when python is shutting down).

Additionally, there are issues with the multiprocessing resource tracker as well:
```python
/home/pavy/micromamba/envs/phd/lib/python3.12/multiprocessing/resource_tracker.py:254: UserWarning: resource_tracker: There appear to be 16 leaked shared_memory objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
```

**Setup:**
 - Debian on Windows Subsystem for Linux
 - python 3.12.1
 - numpy 1.26.3
 - pynbody 1.5.2

**Additional context**
The first issue seems to be a new issue with python 3.12 (see https://github.com/python/cpython/pull/104826). They started emitting errors when threads are spawned during interpreter finalisation. This error can be somewhat avoided by forcing the garbage collector to clean up the arrays before exiting by adding:
```python
import gc
del snapshot
gc.collect()
```
before main() exits.

From what I read from the pynbody code is that these errors come from using the cleanup_callback approach to close and unlink the shared memory?

[...]

PR #790, opened 2024-03-23
Revert back to using posix_ipc module for shared memory support

apontzen:
Addresses #762

### DIFF
--- a/pynbody/array/shared.py
+++ b/pynbody/array/shared.py
@@ -29,8 +29,8 @@ class SharedMemorySimArray(SimArray):
         global _owned_shared_memory_names
         if hasattr(self, '_shared_fname'):
             if self._shared_fname in _owned_shared_memory_names:
-                posix_ipc.unlink_shared_memory(self._shared_fname)
                 _owned_shared_memory_names.remove(self._shared_fname)
+                posix_ipc.unlink_shared_memory(self._shared_fname)
 
 def make_shared_array(dims, dtype, zeros=False, fname=None, create=True,
                       offset = None, strides = None) -> SharedMemorySimArray:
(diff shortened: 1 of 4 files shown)


pavyamsiri:
This PR seems to have fixed it. I don't get any errors/warnings any more.

This was done by reverting back to posix_ipc. Do you think we will go back to using python's native shared memory?
I'm asking because posix_ipc relies on posix and so getting pynbody to work on Windows would be harder.

I have been wanting to get pynbody to work on native Windows for a while but I had to give up and use WSL instead.
This isn't a big deal though I just wanted to ask about your thoughts on reverting.

apontzen:
Thanks, that's good. I'll merge this soon.

Yes part of the logic for using python's own shared memory facilities was exactly because relying on `posix_ipc` is the single most significant block to working natively on Windows. However, after a lot of thought, I think python's own shared memory facilities are a terrible mess, with a misplaced emphasis on safety. Getting it to work _at all_ was cutting against the grain. It's not even clear to me that python 3.13's changes (while welcome) will really fix this.  E.g. As far as I can see it will still refuse to close a shared memory handle when exported pointers exist — this interacts _horribly_ with the unpredictability of python garbage collection, and seems misplaced because POSIX standard explicitly states: _'The mmap() function adds an extra reference to the file associated with the file descriptor fildes which is not removed by a subsequent [close()](https://pubs.opengroup.org/onlinepubs/7908799/xsh/close.html) on that file descriptor. This reference is removed when there are no more mappings to the file.'_

I have isolated the `posix_ipc` calls to a few critical locations, so in principle one could factor these out into a small platform module, with alternative win32 implementations. I don't think this would be a lot of work, but it's also not high on my priority list I'm afraid. If it's something you could take on, I'd be delighted 😄

PR #815, opened 2024-05-08
Fix a couple of issues with unusual AHF catalogues

apontzen:
Where AHF catalogues are either empty or very small (with no substructure), loading them could fail


apontzen:
Now fixes #819

Issue #802, opened 2024-04-21
Documentation updates


apontzen:
Updating all documentation ahead of pynbody v2 release.

#785 has addressed some of these, and others are ongoing

Tutorials etc
------------

- [x] Quick-start
- [x] Reading data
- [x] Filtering data
- [x] Linking snapshots
- [x] Profiles
- [x] Rotation curves
- [ ] Images
- [x] Halos
- [ ] HMF
- [ ] Performance
- [ ] Configuration
- [ ] Threading etc
- [x] Derived quantities

Reference documentation
----------

- [ ] analysis
  - [x] angmom
  - [x] cosmology
  - [x] decomp
  - [ ] gravity
  - [x] halo
  - [ ] hifrac
  - [ ] hmf
  - [ ] interpolate
  - [ ] ionfrac
  - [ ] luminosity
  - [ ] pkdgrav_cosmo
  - [ ] profile
  - [ ] ramses_util
  - [ ] theoretical_profiles
- [ ] array
- [x] bridge
- [x] chunk
- [x] filt
- [x] gravity
- [x] halo
- [x] kdtree
- [x] openmp
- [ ] plot
   - [ ] gas
   - [ ] generic
   - [ ] metals
   - [ ] profile
   - [ ] sph
   - [ ] stars
   - [ ] util
- [x] snapshot
- [ ] sph
- [ ] util
- [ ] configuration
- [ ] dependencytracker
- [ ] derived
- [x] family
- [ ] simdict
- [x] transformation
- [ ] units

apontzen:
Work is now complete

Issue #799, opened 2024-04-14
Geometric filters use periodic BCs even if the snapshot is rotated


apontzen:
Geometric filters wrap particles around the box using simple logic here:

https://github.com/pynbody/pynbody/blob/master/pynbody/filt/geometry_selection.hpp#L11

However, if the snapshot has been rotated so that the box axes don’t align with the coordinate axes, this will give wrong results.

In the same way that the KDTree now disables periodicity on a rotated box, so too should these filters.

A unified approach to this would be best, setting/unsetting a flag that tracks when a simulation is in its original orientation.

Issue #770, opened 2024-02-19
Long-term KDTree enhancements


apontzen:
Recent work on the KDTree identified the following long-term enhancements:

- [ ] Improve code style and C++ semantics, e.g. using functional programming rather than macros and use STL to replace the manual priority queue
- [ ] Implement cuboid queries directly rather than using a bounding sphere
- [ ] Further improve speed of sphere queries by including entire KDTree cells if they are entirely contained by the sphere
- [ ] Add a concept of a 'rotated' and 'translated' KDTree, and construct such a thing when corresponding transformations are performed on the snapshot (for efficiency, and to avoid bugs like #714)

PR #818, opened 2024-05-10
Add a failing test which illustrates the problem with transformations and lazy-loading

apontzen:
The tests will fail on this pull request, intentionally, since this illustrates a test that probably should pass but doesn't because transformations only affect loaded data, not data that later gets lazy-loaded.

This issue is a bit of a corner-case since not so many users have lots of 3D arrays. But it does affect e.g. anyone with magnetic fields.

See also the brief note of this in issue #802, and related commentary in #770, #799


apontzen:
Fixed in #860

PR #821, opened 2024-05-20
AdaptaHOP with long ints

Martin-Rey:
Hi Andrew,

This PR adds a new AdaptaHOP class tailored for versions ran with self-consistent long ints across the code. 

I have tried to avoid duplicating code while making the new loader as backwards compatible with previous AdaptaHOP versions that (i) have iords on short ints; and (ii) have iords on long ints, but the rest of properties and headers in short int (the default previously). 

I have confirmed that this loads the data correctly on an in-house example, and maybe @cphyc can confirm that pynbody can still load older AdaptaHOP catalogues without issues?

Let me know if anything stands out
Martin

### DIFF
--- a/pynbody/halo/adaptahop.py
+++ b/pynbody/halo/adaptahop.py
@@ -208,10 +208,6 @@ class BaseAdaptaHOPCatalogue(HaloCatalogue):
             if self._read_contamination:
                 Nskip += len(self._halo_attributes_contam)
 
-            # We might need a reference to both int readers later in the loop
-            default_reader = fpu.read_int64
-            backup_reader = fpu.read_int
-
             for i in range(nhalos + nsubs):
                 self._file_offsets[i] = fpu.tell()
                 if self._longint:
@@ -220,16 +216,7 @@ class BaseAdaptaHOPCatalogue(HaloCatalogue):
                     npart = fpu.read_int()
                 self._npart[i] = npart
                 fpu.skip(1)  # skip over fortran field with ids of parts
-
-                # AdaptaHOP might use long ints for iord only, or for every
-                # int in the code. Hide this subtlety to the user with a try catch.
-                try:
-                    ipos = fpu.tell()
-                    self._halo_numbers[i] = default_reader()
-                except ValueError:
-                    fpu.seek(ipos)
-                    self._halo_numbers[i] = backup_reader()
-                    backup_reader, default_reader = default_reader, backup_reader
+                self._halo_numbers[i] = self._read_int_without_knowing_length(fpu)
                 fpu.skip(Nskip)     # skip over attributes
 
     def _get_all_particle_indices(self):
@@ -312,15 +299,7 @@ class BaseAdaptaHOPCatalogue(HaloCatalogue):
                 npart = fpu.read_int()
 
             fpu.skip(1) # iord array
-
-            # Same as above -- after PR#821, AdaptaHOP catalogues can have mixed short/long int, or full long ints.
-            try:
-                ipos = fpu.tell()
-                halo_id_read = fpu.read_int64()
-            except ValueError:
-                fpu.seek(ipos)
-                halo_id_read = fpu.read_int()
-
+            halo_id_read = self._read_int_without_knowing_length(fpu)
             assert i == halo_id_read
             if self._read_contamination:
                 attrs = self._halo_attributes + self._halo_attributes_contam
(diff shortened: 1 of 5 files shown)


pynbody/halo/adaptahop.py
```diff
@@ -216,8 +216,11 @@ def _get_halo_numbers_and_file_offsets(self):
                     npart = fpu.read_int()
                 self._npart[i] = npart
                 fpu.skip(1)  # skip over fortran field with ids of parts
-                self._halo_numbers[i] = fpu.read_int()
-                fpu.skip(Nskip) # skip over attributes
+                if self._longint:
+                    self._halo_numbers[i] = fpu.read_int64()
+                else:
+                    self._halo_numbers[i] = fpu.read_int()
```
cphyc, line 222: This is the culprit for the failing tests. I can suggest the following:
```python
            # Initialize this outside the for loop
            default_reader = fpu.read_int64
            backup_reader = fpu.read_int
                
            for i in range(nhalos + nsubs):
                [...]
                try:
                    ipos = fpu.tell()
                    self._halo_numbers[i] = default_reader()
                except ValueError:
                    fpu.seek(ipos)
                    self._halo_numbers[i] = backup_reader()
                    backup_reader, default_reader = default_reader, backup_reader
```

pynbody/halo/adaptahop.py
```diff
@@ -300,7 +303,10 @@ def get_properties_one_halo(self, i):
 
             fpu.skip(1) # iord array
 
-            halo_id_read = fpu.read_int()
+            if self._longint:
+                halo_id_read = fpu.read_int64()
+            else:
+                halo_id_read = fpu.read_int()
```
cphyc, line 309:
```suggestion
            try:
                ipos = fpu.tell()
                self._halo_numbers[i] = fpu.read_int64()
            except ValueError:
                fpu.seek(ipos)
                self._halo_numbers[i] = fpu.read_int()
```

pynbody/halo/adaptahop.py
```diff
@@ -358,7 +364,9 @@ def _detect_longint(cls, fpu: FortranFile, longint_flags: Sequence) -> bool:
             except (ValueError, OSError):
                 pass
 
-        raise ValueError("Could not detect longint")
+        raise ValueError(cls.__name__ + " could not detect longint. "
+                                        "Most likely, this class is expecting the wrong header/data blocks "
+                                        "compared to what is stored in the halo catalogue.")
```
cphyc, line 369:
```suggestion
        raise ValueError(
            f"{cls.__name__} could not detect longint. "
            "Most likely, this class is expecting the wrong header/data blocks "
            "compared to what is stored in the halo catalogue."
        )
```

Martin-Rey:
Thanks @cphyc -- very useful. I believe your solutions now keep the code backward compatible.

The failing tests are now due to me adding a test with this new catalogue, for which the data is missing from the online arxiv. @apontzen would you recommend an extra file to the test archive (210 Mb)?

pynbody/halo/adaptahop.py
```diff
@@ -216,8 +220,17 @@ def _get_halo_numbers_and_file_offsets(self):
...
                 self._npart[i] = npart
                 fpu.skip(1)  # skip over fortran field with ids of parts
-                self._halo_numbers[i] = fpu.read_int()
-                fpu.skip(Nskip) # skip over attributes
+
+                # AdaptaHOP might use long ints for iord only, or for every
+                # int in the code. Hide this subtlety to the user with a try catch.
+                try:
+                    ipos = fpu.tell()
+                    self._halo_numbers[i] = default_reader()
```
apontzen, line 228: Would it not be better here, and in the similar code below, to implement a `read_int_or_int64` method in `FortranFile` that inspects the field size and factors all this away where it more naturally belongs?

apontzen:
I have made a small suggestion for simplifying this a little. One might also consider whether the fortran reader could more generally take alternative dtypes, so that the different header types could be presented to it as alternatives. Then one probably wouldn't even need the slightly awkward `NewAdaptaHOPCatalogueFullyLongInts` class - `NewAdaptaHopCatalogue` could just know there are a couple of alternative header dtypes

apontzen:
Unfortunately to get github to access the latest testdata you need to manually bump the key on line 37 on `build-test.yaml` (I know that's ridiculous, just haven't had a moment to sort it out more methodically)

Martin-Rey:
@apontzen -- with @cphyc, we implemented a cleaner solution for the int reading as you suggested.

And although it would be great to have a more flexible reader that guesses types rather than relying on hardcoded and copy-pasted tuples, it would require a bit more engineering in the `read_attrs` function and significant testing. More flexibility might introduce silent fails and garbage fails as only the length of the record is available for Fortran records (so the type -- integer or float --  and the number of fields expected would still need to be hardcoded somewhere). I think the is beyond what I have time for right now.

PR #823, opened 2024-05-22
NumPy 2.0 support

jobovy:
While trying to determine whether `galpy` runs under `numpy` 2.0, I found that pynbody doesn't compile or run under `numpy` 2.0, so I started trying to fix that. So far I've been able to get the code to compile using `numpy` 2.0 (whose C API can be used with `numpy` <2), but there are still a bunch of errors in the tests that I need to chase down, so a draft PR for now (also need to reconsider all the changes to the C code, because I'm not sure they are all optimal).


jobovy:
Okay, I've now managed to get almost all tests to pass both for `numpy` < 2 and `numpy`==2. However, there are a few tests that fail because of data-type issues within the test data and I've had to make some changes to the code that I don't entirely understand because I'm not that familiar with overriding `numpy.ndarray`, so you'll want to give this a good review. Here are some comments on the changes:
* The most substantial and confusing changes are in `pynbody/array/__init__.py` where I've had to change how `numpy.ndarray` is overridden in the `SimArray` object. But the changes are highly confusing:
  *  Uncontroversially, `__array_wrap__` needs to now take a `return_scalar=` keyword and because I think you probably never return a scalar, this can just be ignored (this is what's done in `astropy` for Quantities; see [here](https://github.com/astropy/astropy/pull/15925/files#r1462645043))
  *   Similarly, `__array__` needs a `copy=None` keyword now that can also, I think, be ignored.
  *   But things got a bit hairy for the `__imul__` and similar functions. I don't quite understand how these work and how they interact with the overridden ufuncs like `multiply`, but with the existing implementation, I was getting weird errors for how the units were, e.g., multiplied when using `__imul__`, where units are multiplied using the `_mul_units` function. This failed under `numpy` 2.0 because the context for `__imul__` that is passed into `__array_wrap__` actually seems to have three arguments rather than the two that are expected for `_mul_units`: (i) the original array, (ii) the array to multiply with, and (iii) the original array again! As best as I can tell, this was also the case for `numpy` < 2, but somehow the code did work in that case (regular `__mul__` has a context that has just two arguments, the arrays to be multiplied). My best guess for what was happening before is that the code wasn't actually working and that that's related to why you had to manually multiply in the units after doing `np.ndarray.__imul__` in the `__imul__` function; I've now had to remove this to not end up with the wrong units.
     
         Having three arguments in the `__array_wrap__` context seems to happen for other internal functions as well (like `__idiv__`, `__iadd__`, and `__isubtract__`). I assume this has to do with these having `self` as an argument, but I can find preciously little documentation on what context to expect for different `numpy` ufuncs. Anyway, to deal with the fact that sometimes there is a third argument that can be ignored, I've just added a dummy keyword to `_mul_units`, `__div_units`, and `_consistent_units` to just be able to be called with three arguments. At least all the tests pass then, but I don't know how complete the tests are in this area.

         I think you'll want to review this carefully because it might have unintended consequences if the tests don't cover this code well.
   * I also have a weird issue in `__array_wrap__` where the line `n_array.sim = self.sim` seems to set `n_array.sim` to None, even though `self.sim` is the correct `SimSnap` and `n_array.sim` is already equal to `self.sim` before this line. So I've just commented out that line and perhaps it should just be removed (I commented it out to remember why I did that).
* Wherever you `cimport numpy` in the Cython code, I've added `np.import_array()` (and similar), because it seems necessary in at least some cases and it's [what's recommended by Cython](https://cython.readthedocs.io/en/latest/src/tutorial/numpy.html#:~:text=It%27s%20necessary%20to%20call%20%22import_array%22%20if%20you%20use%20any%20part%20of%20the%0A%23%20numpy%20PyArray_*%20API.%20From%20Cython%203%2C%20accessing%20attributes%20like%0A%23%20%22.shape%22%20on%20a%20typed%20Numpy%20array%20use%20this%20API.%20Therefore%20we%20recommend%0A%23%20always%20calling%20%22import_array%22%20whenever%20you%20%22cimport%20numpy%22);
* There were a few instances in the code where I had to explicitly cast arrays as `SimArrays`. E.g., in the `hmf.py` code and the `luminosity.py` code. I don't understand why this wasn't necessary before, because as far as I can tell the arrays in question shouldn't have been `SimArrays` before, but either way this shouldn't do any harm
* Relatedly, I had to change how the return value is handled in the direct n-body gravity calculation, by explicitly instantiating a `SimArray`. Again, the code is equivalent, so this shouldn't be an issue. I think this one is necessary in `numpy` 2.0 because of how data-type promotion has changed in `numpy` 2.0, but I'm not sure.
* Because of changes to data-type promotion, I also had to explicitly cast some numbers to `int` or `uint64` (e.g., in the `gadget.py` code).
* I had to do some casting of the `h5py` Datasets in `subfindhdf.py` because `h5py` doesn't itself support `numpy` 2.0, so this should just be a work-around until they do
* Finally, I had to label a bunch of tests as `xfail` because I think they can't be made to work without substantial and uninteresting changes to the tests/code. These are tests involving `testdata/gadget2/test_g2_snap.0`, where the issue seems to be that `pos` is a `float32`, while `mass` and `eps` are `double` and because of changes to numpy's data-type promotion, this means that these tests run into failures. The easiest fix of this might be to change the test data to have consistent data types, unless that's a real use case.
   * Other `xfail` for the test involving the `extinction` package in `tests/sph_image_test.py`, because `extinction` isn't compatible with `numpy` 2.0 yet (I'm not sure it's maintained, I think the last commit is years ago).
   * Other `xfail` for a test in `tests/array_test.py` that seems to involve creating a `Unit` object with 10 scales, which I don't think is really intended usage of this? This seems to have worked fine before, but doesn't any more. But I'm not sure it should be supported as this test seems to have been added for a very specific issue a long time ago.

I've also had to do damage to the `.github/workflows/build-test.yaml` file to get it to work for the `numpy` 2.0 pre-release, all of that could be removed once `numpy` releases 2.0.

Also got rid of `oldest_supported_numpy` because that's deprecated and packages built against 2.0 work with `numpy` 1.x versions (at least recent ones). So the wheels created would be compatible with both `numpy` 1.x and 2.0. BTW, I also noticed that your `numpy~=1.22.4` tests don't actually use 1.22.4, but actually just use 1.26, so those tests are redundant currently. I don't know why this is, but I think some dependency must require a higher numpy and so the 1.22.4 version gets uninstalled and replaced. Anyway, I don't think there's any reason to support 1.22.4 at this point, so you might want to drop that test.

I've cleaned up the commit history, so this could in principle be merged rather than squashed, which might be good to preserve some of the history of the changes.

Apologies for the long comment!

apontzen:
Thanks very much for all this work Jo! I’ll take a careful look soon. One immediate comment is that I’m afraid that people really do use mixed data type files all the time (it’s a cause of endless headaches) so we can’t revoke support for those.

jobovy:
> Thanks very much for all this work Jo! I’ll take a careful look soon. One immediate comment is that I’m afraid that people really do use mixed data type files all the time (it’s a cause of endless headaches) so we can’t revoke support for those.

Okay, I can take another look then, but maybe I'll wait until you've done a more complete review first.

jobovy:
Just a quick note that `numpy` 2.0 is now supposed to be released on June 16: https://numpy.org/news/#numpy-20-release-date-june-16. Since the current version of `pynbody` is clearly not compatible with `numpy` 2.0 and because resolving that incompatibility might take longer than a few weeks, you might want to release a version with a `numpy<2` pin to avoid people using an incompatible `numpy` version. The easiest way to do this might just be a quick post-release `v1.6.0.post0` update of the latest release.

apontzen:
Yes, I’ve pushed out a `1.6.0.post0` as you suggested - hopefully this is helpful downstream until pynbody v2 is released.

apontzen:
I've fixed the remaining tests (except for the one requiring the extinction package), and also substantially simplified the way in which the units are assigned to fully embrace the new mechanisms available. This removes the need for all the confusing `__imul__` overrides etc. Hopefully the new code is a bit more transparent (though in honesty, the underlying numpy interface is just pretty confusing!)

Could you clarify a bit more what's happening inside `subfindhdf.py`? I couldn't quite work out why this data casting should be necessary or how that interacts with h5py's own numpy compatibility

jobovy:
> I've fixed the remaining tests (except for the one requiring the extinction package), and also substantially simplified the way in which the units are assigned to fully embrace the new mechanisms available. This removes the need for all the confusing `__imul__` overrides etc. Hopefully the new code is a bit more transparent (though in honesty, the underlying numpy interface is just pretty confusing!)

Yeah, that all looks a lot cleaner! Thanks for fixing the type incompatibility in the gadget reader.

> 
> Could you clarify a bit more what's happening inside `subfindhdf.py`? I couldn't quite work out why this data casting should be necessary or how that interacts with h5py's own numpy compatibility

The current code gives failures such as
```
DeprecationWarning: __array__ implementation doesn't accept a copy keyword, so passing copy=False failed. __array__ must implement 'dtype' and 'copy' keyword arguments.
```
which is because `h5py` hasn't made the necessary change to accept the `copy=` keyword in `__array__` (like I added in this PR). Somehow this is avoided by doing the explicit cast. What I would do is keep my changes for now (which are obviously okay in terms of functionality) and open an issue to change them back to the simpler version once `h5py` adds numpy 2.0 compatibility.

apontzen:
Done, thanks very much Jo!

Issue #826, opened 2024-05-30
"tcoolform" KeyError when loading a starlog file with a "tCoolForm" entry


trquinn:
**Describe the bug**
Certain ChaNGa versions included "tCoolForm" in the starlog records.  When reading in such starlogs (with the suitable metadata in the log file), a KeyError is raised:
```
Traceback (most recent call last):
  File "/home/u14266/.local/lib/python3.8/site-packages/pynbody/snapshot/__init__.py", line 1932, in _get_array
    return SubSnap._get_array(self, name, index, always_writable)
  File "/home/u14266/.local/lib/python3.8/site-packages/pynbody/snapshot/__init__.py", line 1675, in _get_array
    ret = self.base._get_array(name, util.concatenate_indexing(
  File "/home/u14266/.local/lib/python3.8/site-packages/pynbody/snapshot/__init__.py", line 1209, in _get_array
    x = self._arrays[name]
KeyError: 'tcoolform'

**To Reproduce**
This was was encountered when analysing "near-mint Sandra", run with changa_uw version 6a51ac8.  The log file had to be edited to include the following metadata:
```
# starlog data:
# iOrdStar i4
# iOrdGas i4
# timeForm f8
# rForm[0] f8
# rForm[1] f8
# rForm[2] f8
# vForm[0] f8
# vForm[1] f8
# vForm[2] f8
# massForm f8
# rhoForm f8
# TForm f8
# tCoolForm f8
# H2FracForm f8
# end starlog data
```
The error is then reproduced with
```
s = pynbody.load("h329.cosmo50PLK.3072gst5HbwK1BH.004096")
s.read_starlog()
```

**Expected behaviour or result**
Having the above commands execute without an error being raised.

**Setup (please complete the following information):**
 - OS: Rocky Linux release 8.9 (Green Obsidian)
 - Python version: 3.8.12
 - Numpy version: 1.22.4
 - Pynbody version: 1.4.2


**Additional context**
Data files available on request.

PR #827, opened 2024-05-30
Trq/starlog tcoolform

trquinn:
This is an attempt to fix issue #826 
It creates the appropriate array if tcoolform is detected in the starlog metadata.


trquinn:
I just noticed that this might address the same issue as PR #491

PR #830, opened 2024-06-11
Added Functionality to halo_shape function to perform analysis on stellar and gas shapes, not just dark matter. 

Bvogel4:
Add parameter matter_type to store which type of particle to perform analysis on, defaults to dark matter so existing code should be backwards compatible. 

Rewrote lambda functions to increase readability, inline with PEP E731.

### DIFF
--- a/pynbody/analysis/halo.py
+++ b/pynbody/analysis/halo.py
@@ -456,8 +456,7 @@ def halo_shape(sim, N=100, rin=None, rout=None, bins='equal', use_family='dm'):
         with logarithmic and linear radial spacing.
 
     use_family : str or pynbody.family.Family
-        The type of particles to consider. Default is 'DM'. Example of other str options would be 'Star', 'Gas',
-        Alternatively use pynbody.family.Family object.
+        The particle family to use. Default is 'dm'. 
 
     Returns
     -------
--- a/pynbody/analysis/halo.py
+++ b/pynbody/analysis/halo.py
@@ -477,6 +477,7 @@ def halo_shape(sim, N=100, rin=None, rout=None, bins='equal', use_family='dm'):
         The rotation matrices for each shell.
 
     """
+    from pynbody import family
 
     #-----------------------------FUNCTIONS-----------------------------
     # Define an ellipsoid shell with lengths a,b,c and orientation E:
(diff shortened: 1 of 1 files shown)


pynbody/analysis/halo.py
```diff
@@ -481,17 +484,28 @@ def Ellipsoid(r, a,b,c, E):
         x,y,z = np.dot(np.transpose(E),[r[:,0],r[:,1],r[:,2]])
         return (x/a)**2 + (y/b)**2 + (z/c)**2
 
+
     # Define moment of inertia tensor:
-    MoI = lambda r,m: np.array([[np.sum(m*r[:,i]*r[:,j]) for j in range(3)]\
-                               for i in range(3)])
+    def MoI(r,m):
+        ShapeTensor = np.zeros((3,3))
```
apontzen, line 490: ideally please use `shape_tensor` as per PEP8

pynbody/analysis/halo.py
```diff
@@ -455,6 +455,9 @@ def halo_shape(sim, N=100, rin=None, rout=None, bins='equal', matter_type = 'DM'
         number is not necessarily maintained during fitting. 'log' and 'lin' initialise bins
         with logarithmic and linear radial spacing.
 
+    matter_type : str
```
apontzen, line 458: This should be named something like `use_family`, accepting type of either `str` or `pynbody.family.Family`. (See [here](https://pynbody.readthedocs.io/latest/reference/_autosummary/pynbody.family.html) for background on families.)

See below for how to implement this

pynbody/analysis/halo.py
```diff
@@ -481,42 +484,62 @@ def Ellipsoid(r, a,b,c, E):
...
     # Retrieves alignment angle:
-    almnt = lambda E: np.arccos(np.dot(np.dot(E,[1.,0.,0.]),[1.,0.,0.]))
+    def almnt(E):
+        theta = np.arccos(np.dot(np.dot(E,[1.,0.,0.]),[1.,0.,0.]))
+        return theta
     #-----------------------------FUNCTIONS-----------------------------
 
-    if (rout is None): rout = sim.dm['r'].max()
-    if (rin is None): rin = rout/1E3
+    #select particle type
```
apontzen, line 511: Using the standard pynbody family mechanism (see comment about call signature above), this would be something along the lines of:

`sim_m = sim[pynbody.family.get_family(use_family)]`

This is then completely general if people are using different family names

apontzen (changes requested):
Thank you for your pull request! I have added some small suggestions.

pynbody/analysis/halo.py
Bvogel4: Thank you for the feedback!

I have committed changes to use your suggestions.

apontzen:
pre-commit.ci autofix

apontzen (approved):

apontzen:
@morkney has pointed out that #782 predates this and for some reason hasn't been merged. I'll look at merging that first before coming back to this

PR #832, opened 2024-06-20
New halo shape function fixes

apontzen:
Supplants #782


apontzen:
pre-commit.ci autofix

apontzen:
@morkney @bvogel4 this is my attempt to fix everything in #782 and make it mergeable - could you check it over please? Especially the new backward-compatibility wrapper.

Bvogel4:
Thanks! @apontzen

I have reviewed the changes and tested them with my existing code.

- The wrapper function performs exactly the same as before and provides the deprecation warning as expected.
- I updated the stellar section of my code to call the new `shape` function, and I can confirm that it produced the same output as my custom function.

PR #833, opened 2024-06-26
Simple fix for misleading HBT+ exception message

apontzen:


apontzen:
pre-commit.ci autofix

PR #839, opened 2024-07-05
Update subfindhdf.py

gandhalij:
Redefined creation of offset tables for TNG files only, to account for

1. Group catalogue files broken up into chunks (previously offsets were being reset for each chunk file)
2. Multiple 'PartType's mapped onto the same familiy (previously the offsets were relative to start of the given 'PartType' block)
3. 'Fluff' particles at the end of each FOF group, which are not associated with any Subfind subhalo


apontzen:
pre-commit.ci autofix

apontzen:
Superceded by #843 

I have looked carefully through this and realised it's not really a TNG-specific issue, so have ported the code back
into the main SubfindHDF implementation. In doing so I found a few ancillary issues (e.g. treatment of boundary particles).

Let's continue development over in #843.

PR #840, opened 2024-07-05
 Add support for tipsy snapshots with > 2^32 particles

bwkeller:
pkdgrav3 offers an option to use tipsy snapshots with greater than 2^32 particles.  This limit normally comes from the 32-bit integers in the tipsy header that store the number of particles.  To beat this limit, pkdgrav3 can use the 4-byte header to raise the maximum number of each count to 2^40 (they become 40-bit unsigned integers effectively).


apontzen:
Thanks!

PR #841, opened 2024-07-05
Fixed bugs in the Kennicutt-Schmidt law plotting function

bwkeller:
The plot.stars.schmidtlaw function has two bugs.  

The first is in the radial binning.  Because rmax and rmin are not explicitly passed to profile.Profile, if the young stars do not cover the same radial range as the gas particles, two profiles with very different bin positions will be created, and the gas surface density will not be calculated at the same radial position as the star formation rate surface density.

The second bug is in the calculation of the star formation rate surface density: it is presumed that the units for pretime will be in Myr, and then this value is divided by 1e6, rather than using the robust unit-conversion framework already in pynbody.  I have switched to using this rather than hand-coding unit conversions.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -21,7 +21,7 @@ build:
        - mkdir $READTHEDOCS_OUTPUT/html/tutorials/plots/
        - cp -r docs/*.png $READTHEDOCS_OUTPUT/html/tutorials/plots/
        # totally unclear why the above should be necessary - in my own setup, sphinx puts everything in sensible places
-       - find $READTHEDOCS_OUTPUT/html  # for debugging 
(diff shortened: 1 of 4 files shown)


apontzen:
Thanks Ben - could we add a test that checks the results from this routine?

bwkeller:
I just pushed a test for this, also caught a bug in my implementation!  The old version of the code will fail this test, but the new changes to fix the radius of the profiles make the code pass..

apontzen:
Thanks! I suggest moving make_disc into pynbody.test_utils. Then it looks like this is ready!

bwkeller:
Done!

apontzen:
Thanks!

PR #842, opened 2024-07-07
Phase diagram plot that handles gasoline and ChaNGa superbubble outputs automatically

bwkeller:
I stupidly changed the names of some of the variables output when superbubble is enabled in ChaNGa relative to their original implementation in gasoline.  The code I wrote in `pynbody.plot.gas.rho_T` to handle this was written forgetting this, so I've fixed that now.  The two relevant auxiliary files are:

- `MassHot` in gasoline is `massHot` in ChaNGa
- `Tinc` in gasoline is `tempEff` in ChaNGa.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -21,7 +21,7 @@ build:
        - mkdir $READTHEDOCS_OUTPUT/html/tutorials/plots/
        - cp -r docs/*.png $READTHEDOCS_OUTPUT/html/tutorials/plots/
        # totally unclear why the above should be necessary - in my own setup, sphinx puts everything in sensible places
-       - find $READTHEDOCS_OUTPUT/html  # for debugging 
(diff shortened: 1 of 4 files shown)


apontzen:
Should this perhaps be handled at the level of the name mappers rather than in a single analysis routine?

bwkeller:
I think that does make sense, but I must confess I'm not super familiar with the name mapper architecture.  Would this just consist of a change to default_config.ini?

apontzen:
Yes that's right, just add the appropriate lines there

bwkeller:
I've made the change to use the name mapper instead of hand-fixing `Tinc` and `MassHot`.  I needed to change the tipsy snapshot reader, as it previously had no support for name mapping.  In the interest of not breaking anyone's analysis scripts, you will now have loadable keys for both the gasoline and ChaNGa naming conventions for the two phase properties.

PR #843, opened 2024-07-08
Fix for reading Arepo-SubfindHDF format with multiple files

apontzen:
Supercedes #839


apontzen:
@gandhalij as commented on the other PR, this is a generalised version of your fix.

Could you try it with your setup and check if it works and returns sensible results in the cases where you previously saw problems?

PR #847, opened 2024-09-15
Clean up theoretical_profiles.py and docs

apontzen:
@Martin-Rey seeking your input on this, as part of the clean-up for pynbody v2 (#802).

So far I have made some cosmetic changes:

* Convert `staticmethod` to `classmethod`
* Pull up NFW-only methods to be abstract at base level
* Add docstrings

However, there remains substantial naming convention inconsistency in the interface. Some specific questions:

* The public interface also seems a bit heavy - are all methods needed to be exposed? (E.g. to take one example, is a `log_profile_functional_static` really needed?) 
* Do the class methods (formerly static methods) add anything useful?
* Is `functional` a helpful word in the method names, or could it be removed? 
* What would your thoughts be if I were to rename functions for greater consistency, e.g. `profile_slope` would be more consistent with `profile` than `get_dlogrho_dlogr`?

Thanks for any input! (And let me know if you want to do it yourself rather than have me tinker)

### DIFF
--- a/pynbody/analysis/theoretical_profiles.py
+++ b/pynbody/analysis/theoretical_profiles.py
@@ -1,9 +1,7 @@
 """
+Implements functional forms of known profiles and code to fit them to simulation data.
 
-theoretical_profiles
-====================
-
-Functional forms of common profiles (NFW as an example)
+At present only the NFW profile is implemented, but the code is designed to be easily extensible to other profiles.
 
 """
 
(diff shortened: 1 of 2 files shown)


Martin-Rey:
Hi @apontzen,

All the changes you have made look great!

- Re public interface: I think it can/should be streamlined. The log methods are definitely useful for NFW, but not generally necessary. We can leave it to each class to define them if deemed useful.
- Re class/static methods: I find it useful to access a simple functional form through `NFW_profile.functional(parameters)` without the overhead of initialising a fully-fledged class with these parameters. That was my rational for isolating them as static, but maybe it is more confusing given the abstract object-oriented constructs I went for?
- `functional` can be removed
- Feel free to rename for consistency

apontzen:
@Martin-Rey I'm continuing to make progress with cleaning this, and have another query. Is the ability to instantiate an NFW profile using a three-parameter form (like r_200 + c + r_s) being used anywhere?

It seems like the flexibility introduces a lot of complexity and potential for bugs and I am finding myself drawn to continuing to simplify this class to the absolute bare bones, and only allowing for initialization with r_s + rho_s. This would make implementing e.g. Einasto also trivially easy, which feels like a better direction than having an over-parameterization that may not actually be needed.

I'd really welcome your input on this, especially if you disagree with my thinking!

Martin-Rey:
Hi @apontzen,

I use this functionality pretty regularly in independent analysis code or when visualising profiles. The mass-concentration combination is fundamental to NFW work, while rhos-rs are the direct variables in the functional form, so I find it quite practical to have the option to readily work with either.

I agree that it creates some annoying logic though, and maybe it would be better to strip out this functionality to lie outside of pynbody. Converting between mass and density requires the `_helper_function` and a bit of algebra, which we can leave as a responsibility of the user?

apontzen:
@Martin-Rey would you mind doing a light-touch review over this? I think it's ready to merge...

Martin-Rey (approved):
Looks all good to me, thanks a lot for all the refactoring!

apontzen:
Thanks!

PR #849, opened 2024-09-20
Update to ionisation fraction calculations

apontzen:
As part of #802, this PR updates all ionisation fraction calculations and makes the interface and documentation more robust.

@astroduff this touches your hifrac module (and deprecates it since I think we can now do something more general in the medium term). Please let me know if this will cause issues!


astroduff:
Thanks Andrew, looking forward to seeing the improvement and happy to be deprecated!

PR #864, opened 2024-11-12
Added ipython autocompletion

TomCallingham:
A simple addition to add autocompletion of properties to jupyter/ipython.

Note that this SHOULD be double-checked. It's a small commit and works on my own branch, but I've deviated a fair bit from the main, and it is tricky to swap to the main and sanity check at the moment.

Hopefully, this is more help than hassle! :)


apontzen:
Thanks! Is there a clearly defined way to check that this is working? (I guess one can just plays with tab completion, but would be interested to know what the accepted test is.)

TomCallingham:
I've added a test of sorts, but it does require the Ipython module. I'll leave it to you if you want to include this!

apontzen:
I think this is good, but could you add IPython to the list of test dependences (here: https://github.com/pynbody/pynbody/blob/master/setup.py#L150)?

TomCallingham:
I think that's got it! :)

apontzen:
pre-commit.ci autofix

apontzen:
Thanks!

PR #869, opened 2024-11-20
Update kdmain.cpp

G-Francio:
In case Numpy 2.0 is not available and build isolation fails


apontzen:
Thanks - since numpy 1.x support expires in the [middle of next year](https://scientific-python.org/specs/spec-0000/) I'd be ok with merging this without a formal CI test, provided it works for your use case. What do you think?

PR #873, opened 2024-12-03
Fixed capitalization issue for default band argument in calc_mags

anchwr:
The default band argument for pynbody.analysis.luminosity.calc_mags is 'v', but needs to be updated to 'V' to work with pynbody 2.0.


apontzen:
Thanks!

PR #892, opened 2025-04-24
BUG: Fix out of bounds access in gravity module.

### LINKED ISSUE #891: BUG: Out of bounds access in gravity/_gravity.pyx
**Describe the bug**
In the `_gravity.pyx` cython module, the array `epssq` is indexed by `i` instead of `pi`. This is incorrect as `epssq` has length `nips` and not `n`. This is an out of bounds access which can cause silent crashes.

It does not seem to trigger crashes for snapshots with under 1 million particles but for large enough snapshots this access can lead to crashes.

Aside from crashing, the indexed value would just be incorrect and garbage and so the result would also be wrong.

**To Reproduce**
This bug causes functions like `gravity.midplane_rot_curve` to silently crash python without even a single message.

```python
import pynbody
import numpy as np

import warnings


warnings.filterwarnings("ignore")

# Load data with more than a few million particles
snapshot = pynbody.load("./testdata/gizmo/snapshot_000.hdf5")
print(f"Snapshot has {len(snapshot):,} particles")

# Set softening length
snapshot.properties["eps"] = "0.05 kpc"

# Calculate rotation curve
pos = np.linspace(4, 10, 20)
result = pynbody.gravity.midplane_rot_curve(snapshot.dm, pos)
print(f"Curve = {result}")
print("SUCCESS")
```

The expected result of this script is that `"SUCCESS"` is printed out along with the result of the rotation curve but instead nothing is displayed.

EXPECTED:
```
Snapshot has 12,432,742 particles
Curve = [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
SUCCESS
```

ACTUAL:
```
Snapshot has 12,432,742 particles
```

 - OS: Arch Linux
 - Python version: v3.12.10
 - Numpy version: v2.2.5
 - Pynbody version: v2.1.1

You can access this information from python in the following one-liner:

```python
import sys, platform, pynbody, numpy; print("%s\r\n%s\r\npynbody %s\r\nnumpy %s\r\n"%(sys.version, platform.platform(), pynbody.__version__, numpy.__version__))
```

**Additional context**

pavyamsiri:
Closes #891.

Indexes the `epssq` array using `pi` instead of `i`, fixing an out of bounds array access leading to undefined behaviour.

### DIFF
--- a/pynbody/gravity/_gravity.pyx
+++ b/pynbody/gravity/_gravity.pyx
@@ -74,7 +74,7 @@ def direct(f, np.ndarray[DTYPE_t, ndim=2] ipos, eps=None, int num_threads = 0):
     for pi in prange(nips, nogil=True, schedule='static'):
         for i in range(n):
             mass_i = mass[i]
-            epssq_i = epssq[i]
+            epssq_i = epssq[pi]
             dx = ipos[pi,0] - pos[i,0]
             dy = ipos[pi,1] - pos[i,1]
             dz = ipos[pi,2] - pos[i,2]
--- a/tests/gravity_test.py
+++ b/tests/gravity_test.py
@@ -73,3 +73,19 @@ def test_eps_retrieval_number():
                             -0.06739005, -0.06748439, -0.0695245,
                             -0.06803885, -0.0679833,  -0.07277965, -0.07189107])
     npt.assert_allclose(f['phi'][:10], true_phi_10)
+
+
+@pytest.mark.filterwarnings("ignore:no unit")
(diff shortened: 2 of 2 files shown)


apontzen:
Many thanks. Would you like to add a regression test to gravity_test.py ?

pavyamsiri:
> Many thanks. Would you like to add a regression test to gravity_test.py ?

I have added a test. I wasn't really sure what to assert in the test because the behaviour before is that it segfaults which can't be asserted against but I added one anyway.

Is this okay?

Issue #900, opened 2025-05-30
Potential bug: new luminosity interpolation does not log metallicity


Martin-Rey:
Hi,

When calculating magnitudes, the SSP interpolation function here:https://github.com/pynbody/pynbody/blob/9a785e5d8d98856422bd00b1d106f073c170ce15/pynbody/analysis/luminosity.py#L132
 assumes that ages and metals are given in log10 intervals.

But when called on the stars of a snapshot: here:
https://github.com/pynbody/pynbody/blob/9a785e5d8d98856422bd00b1d106f073c170ce15/pynbody/analysis/luminosity.py#L174
the `metals` array never gets logged:
https://github.com/pynbody/pynbody/blob/9a785e5d8d98856422bd00b1d106f073c170ce15/pynbody/analysis/luminosity.py#L203

This is a reminder to investigate whether this is a bug or if instead the documentation is incorrect.

apontzen:
Fixed

PR #901, opened 2025-06-01
Interpolate metallicities in log space as expected

Martin-Rey:
- Added a test simply testing whether the pynbody-defined table recovers the numbers in the provided SSP file. The main version does not (#900) and the test fails.
- Implement the fix (simply logging the `metals` array as we are logging the ages).

Questions
- In my use cases, the `metals` array always contains non-logged, metal mass fractions. Is this the general behaviour expected for pynbody, or are there snapshot types which output metal mass fractions in log-space already and for which this fix will be breaking? 
- The other luminosity tests are currently failing since the behaviour has changed and numerical values have quantitatively changed. @apontzen I don't have a good understanding of what the "truth" of these tests should be (the big arrays at the top of the file). Maybe it'd be better if you had a look at this part yourself since you coded these tests in #817 ?


apontzen:
All I checked was that the difference from v1 was small and then I hard-coded these values based on the output on my own machine. So if you update to the new values, that will be fine.

It would be useful in doing that to get a sense of how large a difference it makes to some magnitudes (or perhaps we could do this on some science runs instead) which we can add to the docs when releasing this bug fix.

Martin-Rey:
Luminosity tests updated and passing -- is it safe for me to also update the other tests that rely on `luminosity` and are now failing:
```
FAILED plot_stars_test.py::test_sb_profile - AssertionError: 
FAILED sph_image_test.py::test_render_stars - AssertionError: 
FAILED sph_image_test.py::test_render_stars_with_dust - AssertionError: 
FAILED sph_image_test.py::test_render_stars_spherical - AssertionError: 
```

The impact of the error really depends on the combination of ages, metallicities and band considered. For example, it can be catastrophic to assume high-Z in the V band for young (30-Myr) stars when they should be low-Z (factor of 3 difference). But the same comparison for old stars leads to a 10% difference.

As you can see from the code review, the numbers for the integrated halo mag and half-light radius of the Changa test snapshot have changed from 1% in some bands, to 10-20% in others. From comparing between the `v1` and `default` numbers, this looks like a shift of similar magnitude.

apontzen:
Thanks and yes it's fine to update those tests too -- I'd just suggest double checking the values are changing in accordance with the rough magnitude of change you expect.

apontzen:
p.s. If you could also bump the version number up and a suitable `.. versionchanged::` message to the module docstring for `pynbody.analysis.luminosity` with your rough numbers as above, that'd be particularly awesome... I'll copy that text into the release notes too and send out a message on the users mailing list once the fix is on pypi.

Martin-Rey:
@apontzen I believe this is now ready to be merged from a code standpoint, but I am getting a test failure for the docs on `cat docs/conf.py`. First time I see this -- any idea what could be causing it?

Martin-Rey:
Apologies, I thought `version changed` was just a file tag and didn't realise there was a need to describe the changes. Should be done now.

apontzen:
This hasn't quite formatted as expected (https://pynbody--901.org.readthedocs.build/901/reference/_autosummary/pynbody.analysis.luminosity.html)

Also, is mentioning the issue number useful? I don't think so, since the description is self-contained. But if you do think it's helpful, maybe turn it into a link.

Otherwise, we are almost there - many thanks!

Martin-Rey:
Should be better now?

apontzen:
Perfect, many thanks

PR #907, opened 2025-06-12
Gadget hdf snap partial load

wx-ys:
Add partial loading support to GadgetHDFSnap

Implementation:
- Adds a 'take' parameter to GadgetHDFSnap initialization
- Creates local particle type slices based on selected particles
- Updates family slices to reflect the partial loading
- Modifies array loading to selectively read only needed particles

Tests:
- Added test_partial_load to verify functionality

### DIFF
--- a/pynbody/snapshot/gadgethdf.py
+++ b/pynbody/snapshot/gadgethdf.py
@@ -180,10 +180,12 @@ class _HDFFileIterator:
             self.particle_offset = 0
 class _HDFArrayFiller:
     """A helper class to fill a pynbody array from an HDF5 dataset."""
-    
-    def __init__(self):
-        self.sim_element_size = 1 # default element size for simulation arrays
-        self.file_element_size = 1 # default element size for file arrays
+
+    def __init__(self, sim_array_to_fill = None, hdf_dataset = None):
+        
+        # default element size for simulation arrays
+        self.sim_element_size = 1 if sim_array_to_fill is None else self._get_element_size(sim_array_to_fill)
+        self.file_element_size = 1 if hdf_dataset is None else self._get_element_size(hdf_dataset)
         self._update_scaling_factor()
 
     def _update_sim_element_size(self, sim_array_to_fill):
@@ -406,11 +408,9 @@ class HDFArrayLoader:
 
         """
 
-        array_filler = _HDFArrayFiller()
         for loading_fam in all_fams_to_load:
             
-            sim_fam_array = sim[loading_fam][array_name]
-            array_filler._update_sim_element_size(sim_fam_array)
+            sim_fam_array, array_filler = self._get_array_filler(array_name, loading_fam, sim, translated_names)
 
             i0 = 0 # current write position in sim_fam_array
 
@@ -435,7 +435,6 @@ class HDFArrayLoader:
                     if last_file_index != file_iterator.file_index:
                         dataset = self._get_dataset_from_translated_names(sim, file_iterator.current_hdf_file, translated_names)
                         last_file_index = file_iterator.file_index
-                        array_filler._update_file_element_size(dataset)
 
                     if dataset is not None:
                         target_array = sim_fam_array[i0:i1]
@@ -443,7 +442,31 @@ class HDFArrayLoader:
                                                                        offset=file_iterator.particle_offset)
                     file_iterator.particle_offset += readlen
                     i0 = i1
-            
+
+    def _get_array_filler(self, array_name: str, loading_fam: family.Family, sim: SimSnap, translated_names: list[str]):
+        """
+        Set up and return the simulation family array and its corresponding HDF array filler.
+        """
+        
+        sim_fam_array = sim[loading_fam][array_name]
+        
+        # Find the first dataset for this family to determine file element size
+        first_dataset = None
+        for hdf_group_name in self._family_to_group_map[loading_fam]:
+            if self._file_ptype_slice[hdf_group_name].stop <= self._file_ptype_slice[hdf_group_name].start:
+                continue
+            for hdf_group in self._hdf_files.iter_particle_groups_with_name(hdf_group_name):
+                first_dataset = self._get_dataset_from_translated_names(sim, hdf_group, translated_names)
+                if first_dataset is not None:
+                    break
+            if first_dataset is not None:
+                break
+
+        if first_dataset is None:
+            raise KeyError(f"No dataset found for {array_name} in family {loading_fam}")
+        
+        array_filler = _HDFArrayFiller(sim_fam_array, first_dataset)
+        return sim_fam_array, array_filler
 
     def _get_dataset_from_translated_names(self, sim_obj, hdf_particle_group, translated_names):
         """Retrieve an HDF5 dataset from translated_names."""


apontzen:
pre-commit.ci autofix

pynbody/snapshot/gadgethdf.py
```diff
@@ -155,6 +155,200 @@ class _SubfindHdfMultiFileManager(_GadgetHdfMultiFileManager):
...
+            particles_in_range = take[mask]
+
+            if particles_in_range.size > 0:
+                min_id = particles_in_range[0]
+                max_id = particles_in_range[-1]
+
+                # Check if indices form a continuous sequence
+                expected_count = max_id - min_id + 1
+                if particles_in_range.size != expected_count:
+                    raise ValueError(f"Non-continuous indices detected in {ptype_name}. "
```
apontzen, line 238: Could you explain more what this is about? I don't see a test for it and it's not totally clear when it's expected to be raised.

pynbody/snapshot/gadgethdf.py
```diff
@@ -155,6 +155,200 @@ class _SubfindHdfMultiFileManager(_GadgetHdfMultiFileManager):
...
+        
+        for loading_fam in all_fams_to_load:
+            i0 = 0
+            for hdf in sim._all_hdf_groups_in_family(loading_fam):
+                npart = hdf['ParticleIDs'].size
+                if npart == 0:
+                    continue
+                i1 = i0+npart
+
+                for translated_name in translated_names:
```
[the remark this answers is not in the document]
wx-ys: Yes, thanks for pointing this out! I've refactored the duplicated code into two new methods: `_get_dataset_from_translated_names` and `_fill_array_from_hdf_dataset`

apontzen:
This now seems to reimplement considerable parts of the functionality already implemented in `chunk.py`, which was intended to be generic. You can see [the documentation](https://pynbody.readthedocs.io/latest/reference/_autosummary/pynbody.chunk.html) for more info, or [nchilada.py](https://github.com/pynbody/pynbody/blob/master/pynbody/snapshot/nchilada.py) for a reference implementation. 

Would you consider using this approach, to help keep the codebase as unified as possible?

pynbody/snapshot/gadgethdf.py
```diff
@@ -155,6 +155,208 @@ class _SubfindHdfMultiFileManager(_GadgetHdfMultiFileManager):
...
+                            if len(buf_index) > 1 and buf_index[-1] - buf_index[0] == len(buf_index) - 1:
+                                # convert to slice for efficiency if the indices are contiguous
+                                buf_index = slice(buf_index[0], buf_index[-1] + 1)
+
+                        self._fill_array_from_hdf_dataset(target_array, dataset, source_sel=buf_index)
+                    current_hdf_i0 = current_hdf_i0 + readlen
+                    i0 = i1
+            
+
+    def _get_dataset_from_translated_names(self, sim_obj, hdf_particle_group, translated_names):
```
apontzen, line 284: Because this now gets executed every chunk, rather than once per particle type, I wonder whether it might be inefficient? See broader comment below about timing.

pynbody/snapshot/gadgethdf.py
```diff
@@ -155,6 +155,208 @@ class _SubfindHdfMultiFileManager(_GadgetHdfMultiFileManager):
...
+        if isinstance(source_sel, np.ndarray):
+            # Optimization for non-continuous reads:
+            # Read a single contiguous chunk from HDF5 that spans all requested indices
+            # for this file, then select the specific indices from this in-memory chunk.
+            id_min = source_sel[0]
+            id_max = source_sel[-1]
+            num_read = id_max - id_min + 1
+            indices_in_read_chunk = source_sel - id_min
+            if sim_element_size != file_element_size:
+                contiguous_hdf_slice = np.s_[int(id_min * sim_element_size / file_element_size) : int((id_max + 1) * sim_element_size / file_element_size)]
```
apontzen, line 318: This is very cryptic and could do with either a comment or factoring out into a separate method with a helpful name...

pynbody/snapshot/gadgethdf.py
```diff
@@ -272,42 +477,16 @@ def _all_hdf_groups_in_family(self, fam):
...
-            family_slice_start += family_length
+    def __init_file_map(self, take):
+        self._array_loader = HDFArrayLoader(self._hdf_files, self._families_ordered(), self._family_to_group_map, take)
+        self._gadget_ptype_slice = self._array_loader._file_ptype_slice
+        self._family_slice = self._array_loader._family_slice_to_load
+        self._num_particles = self._array_loader._num_particles_to_load
 
-
-        self._num_particles = family_slice_start
+        # PartType3 is Tracer in Arepo, so we'd better to adjust the group mapping, removing empty groups
```
apontzen, line 486: I don't understand this comment. the below seems to check for empty groups and remove them. What does this have to do with tracers specifically?

This manipulation probably also belongs in `__init_family_map` rather than here

apontzen (commented):
Thank you for all your hard work on this, it is looking very good now. I have attached some final comments.

If possible, it would be helpful to compare the time taken to load arrays from a big file with the current main branch, and with this PR (with `take=None` for a fair comparison). If there is a significant performance penalty, profiling to understand where it comes from would be helpful, but at first it would be fine just to check the overall times.

wx-ys:
> Thank you for all your hard work on this, it is looking very good now. I have attached some final comments.
> 
> If possible, it would be helpful to compare the time taken to load arrays from a big file with the current main branch, and with this PR (with `take=None` for a fair comparison). If there is a significant performance penalty, profiling to understand where it comes from would be helpful, but at first it would be fine just to check the overall times.

Thank you for the feedback, I've learned a lot during this process.

As you suggested, I've run a performance comparison between the current main branch and this PR.

Test Setup
The test was performed on a snapshot with approximately 2.7 x 10⁷ particles, which results in ~50 chunks for loading based on the default buffer size. The particle distribution is as follows:
```
{'PartType0': slice(0, 11933766, None),
 'PartType1': slice(11933766, 21610846, None),
 'PartType2': slice(21610846, 24717279, None),
 'PartType3': slice(24717279, 24717279, None),
 'PartType4': slice(24717279, 26661709, None),
 'PartType5': slice(26661709, 26661716, None)}
```
The following script was used to measure the total time for 10 iterations of loading the snapshot and accessing the pos, vel, mass, and iord arrays:
```
def profile_load():
    """Loads the snapshot and accesses several arrays."""
    snap = pynbody.load("/home/yxi/Simulation/sims/TNG50_l7_Halo274_l11/output/snapdir_099/snap_099/")
    _ = snap['pos']
    _ = snap['vel']
    _ = snap['mass']
    _ = snap['iord']

# The test was run for 10 iterations and the total time was measured.
```
Results
Here are the total times recorded for 10 full loads:

main branch: 8.73s
This PR (after 'Optimize by profiling'): 10.03s
This PR (before 'Optimize by profiling'): 13.82s

pynbody/snapshot/gadgethdf.py
wx-ys: It's now moved into a new method _get_contiguous_hdf_slice of a newly added class _HDFArrayFiller

pynbody/snapshot/gadgethdf.py
wx-ys: Because load_array calls it, I changed the logic there. Now it only gets called when the file changes.

pynbody/snapshot/gadgethdf.py
```diff
@@ -155,6 +155,208 @@ class _SubfindHdfMultiFileManager(_GadgetHdfMultiFileManager):
...
+                    if dataset is not None:
+                        target_array = sim[loading_fam][array_name][i0:i1]
+                        if isinstance(buf_index, slice):
+                            buf_index = slice(buf_index.start + current_hdf_i0, buf_index.stop + current_hdf_i0)
+                        if isinstance(buf_index, np.ndarray):
+                            buf_index = buf_index + current_hdf_i0
+                            if len(buf_index) > 1 and buf_index[-1] - buf_index[0] == len(buf_index) - 1:
+                                # convert to slice for efficiency if the indices are contiguous
+                                buf_index = slice(buf_index[0], buf_index[-1] + 1)
+
```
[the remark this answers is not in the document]
wx-ys: _fill_array_from_hdf_dataset has been changed into a new class _HDFArrayFiller, so this has now moved into _HDFArrayFiller.

apontzen:
My main question is now over the new `_HDFArrayFiller` class. It looks a bit weird that this object is kept alive for the whole lifetime of `HDFArrayLoader`, but its internal state is actually only valid while a _single_ array is being loaded. This is a bit confusing and possibly invites bugs later...

Is it possible just to construct it as each array gets loaded, rather than keep it alive in an invalid state in the background?

Re performance optimization, this possibly looks good but I did not quite follow what `chunk_num` refers to in this context? For someone reading the entire file, is it expected to be `num_particles_in_file // max_chunk`, in which case it could get very large indeed?

wx-ys:
You're right! I will create a short-lived variable `array_filler = _HDFArrayFiller()` within the `load_arrays()` method instead.

Regarding the chunk_num: yes, `chunk_num = num_particles_in_file // max_chunk`. In my tests, chunk_num was 50 when using the default `max_chunk = 1024 * 512`. To simulate scenarios with larger chunk_num values and identify potential performance bottlenecks, I manually adjusted `max_chunk = 1024 * 50`, `causing chunk_num ~ 500`.

pynbody/snapshot/gadgethdf.py
```diff
@@ -154,6 +158,302 @@ class _SubfindHdfMultiFileManager(_GadgetHdfMultiFileManager):
...
+        translated_names : list of str
+            A list of possible names for the array in the Gadget HDF file.
+
+        """
+
+        array_filler = _HDFArrayFiller()
+        for loading_fam in all_fams_to_load:
+            
+            sim_fam_array = sim[loading_fam][array_name]
+            array_filler._update_sim_element_size(sim_fam_array)
```
apontzen, line 413: it would seem to make sense to instantiate `_HDFArrayFiller` here, with `sim_fam_array` as an argument to the constructor -- again eliminating the possibility of using it in an uninitialized state

pynbody/snapshot/gadgethdf.py
```diff
@@ -154,6 +158,302 @@ class _SubfindHdfMultiFileManager(_GadgetHdfMultiFileManager):
...
+                        file_iterator.select_file): 
+                    if mem_index is None or file_iterator.current_hdf_file is None:
+                        continue
+                    i1 = i0 + mem_index.stop - mem_index.start
+
+                    # Check if we need to load a new dataset
+                    if last_file_index != file_iterator.file_index:
+                        dataset = self._get_dataset_from_translated_names(sim, file_iterator.current_hdf_file, translated_names)
+                        last_file_index = file_iterator.file_index
+                        array_filler._update_file_element_size(dataset)
```
apontzen, line 438: is there any sensible scenario in which different files in a series would actually present different element sizes? I am not sure why this needs updating once per file. It would be neatest if this is setup when constructing the `_HDFArrayFiller` above, then there is just a single `fill_array_from_hdf_dataset` call for each chunk

wx-ys:
Yes, I'm ready to merge!  Thank you for the thorough review process.

PR #910, opened 2025-06-26
PKDGRAV3 hdf5 snapshots support

isaac-aa:
Hello,

This PR implements the `pkdgravhdf` loader (heavily derived from `gadgethdf`), which is able to read snapshots from the latest versions of [PKDGRAV3](https://pkdgrav3.readthedocs.io/en/latest/) (also presented [here](https://ui.adsabs.harvard.edu/link_gateway/2023MNRAS.519..300A/PUB_HTML)). The most notable difference is the unit system.

I have implemented some basic tests and modified  the documentation in the places  with references to the available loaders (maybe I missed some?).

As the test data is not at Zenodo, tests will fail. I can pass around the test data (~3 MB) in the meantime. 

Thanks!

### DIFF
--- a/pynbody/snapshot/pkdgravhdf.py
+++ b/pynbody/snapshot/pkdgravhdf.py
@@ -27,16 +27,36 @@ for hdf_groups in _pkd_default_type_map.values():
 
 
 class _PkdgravHdfMultiFileManager(_GadgetHdfMultiFileManager) :
-    _multifile_manager_class = _GadgetHdfMultiFileManager
     _nfiles_groupname = "Header"
     _nfiles_attrname = "NumFilesPerSnapshot"
     _namemapper_config_section = "pkdgrav3hdf-name-mapping"
+
+    def __init__(self, filename, mode='r') :
+        filename = str(filename)
+        self._mode = mode
+        if h5py.is_hdf5(filename):
+            self._filenames = [filename]
+            self._numfiles = 1
+        else:
+            h1 = h5py.File(filename + ".0", mode)
+            self._numfiles = self._get_num_files(h1)
+            if hasattr(self._numfiles, "__len__"):
+                assert len(self._numfiles) == 1
+                self._numfiles = self._numfiles[0]
+            self._filenames = [filename + "."
+                               + str(i) for i in range(self._numfiles)]
+
+        self._open_files = {}
 
     def get_cosmo_attrs(self):
         return self[0].parent['Cosmology'].attrs
 
 
 class PkdgravHDFSnap(GadgetHDFSnap):
+    # PKDGRAV3 creates files by default with the format
+    # {name}.{step:05d}.{n}
+    # where `n>=0` is the index of the file when doing parallel writting.
+    # It does not contain any ".hdf5" extension in the file names.
     _multifile_manager_class = _PkdgravHdfMultiFileManager
     _readable_hdf5_test_group = "Header"
     _readable_hdf5_test_attr = "Header", "PKDGRAV version"
@@ -52,6 +72,10 @@ class PkdgravHDFSnap(GadgetHDFSnap):
             for hdf_family_name in _pkd_all_hdf_particle_groups:
                 if hdf_family_name in hdf:
                     yield hdf[hdf_family_name]
+
+    @classmethod
+    def _guess_file_ending(cls, f):
+        return f.with_suffix(f.suffix + ".0")
 
     def _init_unit_information(self):
         try:
--- a/pynbody/snapshot/pkdgravhdf.py
+++ b/pynbody/snapshot/pkdgravhdf.py
@@ -1,0 +1,129 @@
+import configparser
+import logging
+import warnings
+
+from .. import config_parser, family, units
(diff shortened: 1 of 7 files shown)


apontzen:
Thanks, please point me to your test `pkdgrav3.tar.gz` and I will create a new zenodo version

isaac-aa:
You should be able to download it at: https://cloud.iac.es/index.php/s/dGtA2LAiNqRPzjZ

apontzen:
Thanks, this is now uploaded to zenodo and the new identifier is [15769415](https://doi.org/10.5281/zenodo.15769415)

You'll need to update the `test_data_url` in `test_utils` package, and then hopefully the tests will start running...

pynbody/snapshot/pkdgravhdf.py
```diff
@@ -0,0 +1,153 @@
...
+    for hdf_group in hdf_groups:
+        _pkd_all_hdf_particle_groups.append(hdf_group)
+
+
+class _PkdgravHdfMultiFileManager(_GadgetHdfMultiFileManager) :
+    _nfiles_groupname = "Header"
+    _nfiles_attrname = "NumFilesPerSnapshot"
+    _namemapper_config_section = "pkdgrav3hdf-name-mapping"
+
+    def __init__(self, filename, mode='r') :
```
apontzen, line 34: This appears to be largely copy/pasted from _GadgetHdfMultiFileManager, which obscures what is supposed to be different/the same.

As far as I can see it's only the filename that is different.

As such, could you change `_GadgetHdfMultiFileManager` so that the filename is made in a `_make_filename_for_cpu` method, and then you only need to override that one method?

pynbody/snapshot/pkdgravhdf.py
```diff
@@ -0,0 +1,153 @@
...
+class PkdgravHDFSnap(GadgetHDFSnap):
+    # PKDGRAV3 creates files by default with the format
+    # {name}.{step:05d}.{n}
+    # where `n>=0` is the index of the file when doing parallel writting.
+    # It does not contain any ".hdf5" extension in the file names.
+    _multifile_manager_class = _PkdgravHdfMultiFileManager
+    _readable_hdf5_test_group = "Header"
+    _readable_hdf5_test_attr = "Header", "PKDGRAV version"
+
+    def __init__(self, filename, ):
```
apontzen, line 64: Since this just calls `super().__init__`, I think it can be deleted

pynbody/snapshot/pkdgravhdf.py
```diff
@@ -0,0 +1,153 @@
...
+        super().__init__(filename)
+
+    def _get_units_from_hdf_attr(self, hdfattrs) :
```
apontzen, line 67: If you expect this not to be called, it should raise an exception.

If you do expect it to be called, then it should return something (NoUnit or None perhaps -- unsure what the intent is).

pynbody/snapshot/pkdgravhdf.py
```diff
@@ -0,0 +1,153 @@
...
+        pass
+
+    def _all_hdf_groups(self):
```
apontzen, line 70: Is there a reason why this can't be dealt with in `_PkdgravHdfMultifileManager`, by overriding `iter_particle_groups_with_name`? That would seem to be the neatest place to do it, otherwise we risk subtle inconsistencies between different parts of the code.

tests/pkdgravhdf_test.py
```diff
@@ -0,0 +1,69 @@
...
+                       + snap.g['mass'].sum()
+                       + snap.bh['mass'].sum()
+                       )
+
+
+def test_standard_arrays(snap) :
+    """Check that the data loading works"""
+
+    for s in [snap] :
+        s.dm['pos']
```
apontzen, line 57: You should probably select a few of these to check that the values being returned are actually sensible. I realise `gadgethdf_test` doesn't do that, but that is a historic mistake. For an example of checking the explicitly loaded data, see `ramses_test.py`'s `test_all_dm`.

pynbody/snapshot/pkdgravhdf.py
isaac-aa: Yes, it was hanging from a previous implementation. I have removed it in https://github.com/pynbody/pynbody/pull/910/commits/484e21ac7fe6a39f3c38f9c10816546de4cedbf6

tests/pkdgravhdf_test.py
isaac-aa: I have added more tests in https://github.com/pynbody/pynbody/pull/910/commits/0c49a49b6330a8fb9e92d47f19ec3f3e3a66ccca
I think that this cover most of the "typical" operations, but let me know if you require more thorough testing.

pynbody/snapshot/pkdgravhdf.py
apontzen: Great, but isn't this now a copy/paste of what's in gadgethdf.py? I think this override can be removed entirely now.

apontzen: actually I see the distinction. It's a bit messy but that's reflecting a design oddity in `gadgethdf` module, so I don't immediately see a better way to do it.

PR #914, opened 2025-07-02
Add class to enable tracking even with ramses iord int32-truncation bug

apontzen:
Some versions of ramses truncate iord values to int32 when communicating. This introduces a special bridge class that uses heuristics to recreate the correct snapshot-to-snapshot mapping as far as possible.

### DIFF
--- a/tests/bridge_test.py
+++ b/tests/bridge_test.py
@@ -230,3 +230,52 @@ def test_fuzzy_match_halos(snapshot_pair):
     assert fuzzy_match_rev.keys() == set(f2['grp'])
 
     assert fuzzy_match_rev[1] == [(0, 2./3), (1, 1./3)]
+
+def test_ramses_bug_bridge():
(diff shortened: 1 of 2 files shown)


apontzen:
pre-commit.ci autofix

tests/bridge_test.py
```diff
@@ -230,3 +230,50 @@ def test_fuzzy_match_halos(snapshot_pair):
...
+    index = np.concatenate((index_st, index_dm))
+
+    f_sub = f1[index]
+
+    b = pynbody.bridge.RamsesBugOrderBridge(f1, f2)
+
+    bridged_iord = b(f_sub)['iord_no_bug']
+    unbridged_iord = f_sub['iord_no_bug']
+
+    assert len(np.setdiff1d(bridged_iord, unbridged_iord)) == 0
```
apontzen, line 279: This should be a set comparison, not a set diff

PR #923, opened 2025-08-18
Bugfix for angular momentum calculation

### LINKED ISSUE #922: Angular momentum calculation bug
In `pynbody/analysis/angmom.py`'s align method, a check is made on lines 203-208, the set of particles used for centering is chosen based on how many particles of a given type the halo has (greater than 5 for stars or gas).  However, the actual set of particles used for the centre is the filtered set, `h.gas[filt.Sphere(disk_size)]`, which may have zero particles.  I believe lines 203 and 205 should check the length of `h.gas[filt.Sphere(disk_size)]` etc, rather than just `h.gas`.

I'll put in a PR to try fixing this.

bwkeller:
I've changed the checks on lines 203 and 205 to refer to the filtered sets of particles that are used to calculate the angular momentum vector.  This PR should fix #922 .

### DIFF
--- a/tests/transformation_test.py
+++ b/tests/transformation_test.py
@@ -144,7 +144,16 @@ def test_chaining(test_simulation_with_copy):
 
     npt.assert_almost_equal(f['pos'], original['pos'])
 
-def test_angmom_align(test_simulation_hole):
+def test_align_particle_selection(test_simulation_hole):
+    '''
+    Previously, the align() function in the angmom module had a bug (GH Issue #922) where
+    a series of checks intended to calculate the angular momentum vector of the central region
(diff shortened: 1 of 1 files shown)


apontzen:
Thanks Ben - are you able to add a test that captures this issue? I think it'd be feasible to use `pynbody.new` to create a distribution of particles that triggers the bug in the original code version, and pass it in to check everything succeeds.

No idea what is going on with windows + py 3.13, I don't think this is related to your code change but I guess we need to figure it out... re-running doesn't seem to help, so this is not some one-off thing.

apontzen:
I found the issue, it was use of an uninitialized variable (#921). This is now merged, so if you rebase you should find the existing tests pass. Hopefully we can add one more then get this merged too!

PR #924, opened 2025-09-01
Add a class to allow remote access to SWIFT snapshots

jchelly:
This adds a new simulation snapshot class which can read SWIFT snapshots over the network.

The server at https://dataweb.cosma.dur.ac.uk:8443/hdfstream provides access to HDF5 datasets by reading and serializing them in messagepack format. The [hdfstream python module](https://github.com/jchelly/hdfstream-python) can be used to request data using syntax similar to h5py and I've used this to implement a RemoteSwiftSnap class based on the existing SwiftSnap and GadgetHdfSnap classes. For testing purposes I've put a couple of SWIFT example snapshots on the server.

There are a few outstanding issues:

I'm not sure how testing should work. So far I've run tests by hosting the pynbody SWIFT test data on the server and adapting the SwiftSnap tests to work on remote data. This means that the tests will only pass if the server is accessible and serving the expected files. To set up unit tests for the hdfstream module I used pytest-recording, which records http requests and responses during an initial test run so that later tests can be run without network access. We could do the same here, although the response files might be quite large. For hdfstream I just kept the test responses small enough to commit to the repository.

To implement the take_region option to extract regions from a swift snapshot we need to read multiple contiguous dataset slices. The SwiftSnap class does this by creating virtual datasets that reference the required parts of the dataset. But since we can't create a local dataset that references files on the server I've written a class that wraps a dataset and provides a view of a subset of the elements. If the wrapped dataset is on the remote server then the dataset slices corresponding to the selected region are downloaded with a single http request.

The changes here conflicted a bit with #907, which was merged after I wrote most of this. I believe that PR causes read operations to be split into small chunks. I've had to set the maximum chunk size to be very large when reading a remote swift snapshot so that the code doesn't emit huge numbers of http requests.

I haven't made any changes to pynbody.load(). At the moment remote snapshots can be opened with something like:
```
import pynbody.snapshot.remote_swift as rs
snap = rs.RemoteSwiftSnap("SWIFT/examples/EAGLE_low_z/EAGLE_100/eagle_0000.hdf5", server="https://dataweb.cosma.dur.ac.uk:8443/hdfstream")
```
Since there's only one supported remote snapshot format maybe this is ok for now?


apontzen:
@jchelly apologies it has taken so long but I am finally getting to looking at this. 

It's exciting to see that this works!

A few initial thoughts before getting into any detail:

* is there a publicly available server implementation? If so, the best way to test would probably be to get it running on localhost in the CI environment. Otherwise we either need to record some responses (but I suspect that might be hard to maintain), or accept that `dataweb.cosma` will get inundated by many requests just for testing -- this might lead to problems, and a lot of test failures too if `dataweb` rate limits or is otherwise unavailable.
* As you pointed out, #907 implements a common use case where one wants to load only small subsets of the full file, e.g. just one halo, in a way that's only really appropriate for local files. This is on top of the separate region loading mechanism that swift's indexing provides, but highly complementary. In terms of using the current implementation for a remote service, it seems to me like the tradeoff is between number of requests, and wastefulness of transferring data that will be immediately discarded on reception. Is there an accepted model for requesting a subset of particles via hdfstream  and if so should we _disallow_ the local implementation on remote snapshots and point users to an alternative mechanism? (Or, even better, transparently map the user's request in a way different to the local implementation?)
* Overall I'd like to reduce the level of code duplication if possible. I am not sure the best way to do this, but do think it would be worthwhile -- e.g. doesn't feel right to duplicate the set of tests since it'll lead to maintenance trouble later, and I also see quite a bit of duplication between `swift.py` and `remote_swift.py`.

If helpful to have a chat about next steps, please let me know, apologies again this has taken so long to get to.

jchelly:
Thanks for taking a look. I hadn't thought of just running an instance of the server for testing, but it makes sense. I was meaning to tidy up the repository and make it public anyway so I can set up automated building and testing on github. Right now I have unit tests that run during the build but for testing the full system I'm running on Cosma and querying the EAGLE data.

Regarding partial access: the most efficient way to request a region via hdfstream is to generate a list of slices to read and then send that to the server as a single request. The server returns an array with the slices concatenated along the first dimension. Unfortunately there's no equivalent h5py API for that so the code needs to know it's using a remote file. I recently updated the client module so that you can also index a remote dataset with an array of integers or booleans to do partial reads. It translates the index array into a sorted, unique list of slices, requests those slices from the server, then rearranges them back into the requested order. So it might be possible to use that mechanism for partial loading too.

I agree about the code duplication. I was trying not to break any existing code by just adding new classes, but it will be more work to maintain. Would it make sense to only have one swift reader class that does not require using virtual files, so that it can work on local and remote files?

jchelly:
Sorry for the delay in getting to this. I thought Docker wasn't available on Windows github runners, but apparently that's not the case. I'll try to get the tests working on Windows and then look at removing the duplication.

jchelly:
I don't think it will be possible to use the container version of the server on github Windows runners. Docker containers normally share the OS kernel with the host. Docker can run in a mode where the container starts in a VM instead so it can use a different OS, but as far as I can see that's not available on github runners. Making a Windows server image would be tricky too because the server uses some posix features.

We could maybe have a workflow dispatch trigger to test against an external server (e.g. Cosma) so that we can check remote access works on Windows without adding an external dependency to every CI run.

I have an updated version of this branch where the swift tests are repeated for local and remote files using a fixture. I've also removed the duplicated swift snapshot implementation. I'll push it when Cosma is back. I wasn't quite fast enough yesterday.

jchelly:
In another branch (temporary PR at https://github.com/jchelly/pynbody/pull/9) I've tried a different approach to all this:

  * Modify the `SwiftSnap` class so that it uses the `GadgetHDFSnap(take=...)` mechanism to cut out regions. I.e. treat `take_swift_cells` and `take_region` as just alternate ways to parameterize `take`. Then we don't need to create temporary virtual files or use the sliced dataset view class.
  * Modify `GadgetHDFSnap` to access the hdfstream service. The `HDFArrayLoader` and `HDFArrayFiller` classes work on remote files without any changes. Requesting multiple slices (like a region) efficiently can be handled by special casing the `HDFArrayFiller._fill_from_fancy_index` method.
 
This way we get remote access to non-Swift HDF5 snapshots (e.g. EAGLE) for free and I think it feels better to have just one mechanism for reading subsets.

apontzen:
That branch looks like a very good way to  go - do you want me to do a detailed read-through, or wait for you to open a pull request targeting pynbody:master?

jchelly:
It's still work in progress so probably best not to spend any time reading it in detail yet. Perhaps I should make a pull request that just replaces the SwiftSnap implementation first, without the remote access changes?

jchelly:
I've made a new PR for just the SwiftSnap changes in #1003.

jchelly:
I'll close this because I think the approach in #1003 and #1007 is better.

Issue #926, opened 2025-09-04
Loading nchilada snapshots incorrectly handles files with more than 2 billion particles


trquinn:
**Describe the bug**
The NchiladaSnap _load_header() method incorrectly unpacks the nchilada header, ignoring the higher 4 bytes of the nbodies data.

**To Reproduce**
Attempt to load a large nchilada file (with gas and dark matter) with the command:

s = pynbody.load("romulus25.3072g.ad.000032")

It fails with the following error:

  File "/home1/00333/tg456090/miniconda3/lib/python3.13/site-packages/pynbody/sn                                                                                        apshot/__init__.py", line 47, in load                                                                                                                                    
    return c(filename, *args, **kwargs)                                                                                                                                  
  File "/home1/00333/tg456090/miniconda3/lib/python3.13/site-packages/pynbody/sn                                                                                        apshot/nchilada.py", line 104, in __init__                                                                                                                               
    self._decorate()                                                                                                                                            
    ~~~~~~~~~~~~~~^^                                                                                                                                            
  File "/home1/00333/tg456090/miniconda3/lib/python3.13/site-packages/pynbody/sn                                                                                        apshot/simsnap.py", line 1703, in _decorate                                                                                                                              
    fn(self)                                                                                                                    
    ~~^^^^^^                                                                                                                    
  File "/home1/00333/tg456090/miniconda3/lib/python3.13/site-packages/pynbody/sn                                                                                        apshot/tipsy.py", line 1583, in param2units                                                                                                                              
    sim.gas["rho"].units = denunit_st                                                                                                                                    
    ^^^^^^^                                                                                                                 
  File "/home1/00333/tg456090/miniconda3/lib/python3.13/site-packages/pynbody/sn                                                                                        apshot/simsnap.py", line 462, in __getattr__                                                                                                                             
    raise AttributeError("{!r} object has no attribute {!r}".format(                                                                                                     
        type(self).__name__, name))                                                                                                                                      
AttributeError: 'NchiladaSnap' object has no attribute 'gas'                                                                                                            
>>>                                                                                             

Obviously posting an example snapshot would be difficult: the one I'm currently working with is 3TB.

**Expected behaviour or result**

I expect a successful load of the header information such that "s.gas" would produce:
<FamilySubSnap "romulus25.3072g.ad.000032::gas" len=28991029248>

**Setup **
[...]

PR #927, opened 2025-09-04
NchiladaSnap: correct reading of "nbod" in _load_header().

trquinn:
This lets NchiladaSnap handle simulations with larger than 2**31 particles.  It is a proposed fix for #926


apontzen:
Thanks Tom - I am puzzled at why this doesn’t break existing files where it would seem we previously succeeded loading with a int32 nbod field. I am worried that maybe the test suite just misses a problem - could you give any insight?

PR #931, opened 2025-09-12
simulation_halo_mass_function(): avoid reading masses if not needed.

trquinn:
This is important for plotting halo mass functions for large simulations where even reading the masses is not possible because of limited memory.


apontzen:
Thanks!

PR #932, opened 2025-09-14
Fix incorrect gridsize in hist2d when not using KDE

wx-ys:
In the `hist2d` function, when `use_kde=False` (the default value), there was a discrepancy in how the `gridsize` was handled. This issue caused incorrect binning when the x and y had different bin counts.

 This PR fixes this bug, and adds a test case for this.


apontzen:
Thanks!

PR #934, opened 2025-09-15
Fix units for numpy.gradient, as used in kappa profile

wx-ys:
This PR fixes a bug where the `kappa` profile property lost its units due to use of `np.gradient` causing UnitsException. The fix ensures units are preserved. A test is added to verify correct units and behavior.

### DIFF
--- a/pynbody/array/__init__.py
+++ b/pynbody/array/__init__.py
@@ -406,20 +406,21 @@ class SimArray(np.ndarray):
 
         result = func(*args_processed, **kwargs_processed)
 
-        if func in SimArray._ufunc_registry:
-            
-            sim = None
-            for arg in args:
-                if isinstance(arg, SimArray):
-                    sim = arg.sim
-                    break
-            units_val = SimArray._ufunc_registry[func](*args, **kwargs)
+        units_rule = SimArray._ufunc_registry.get(func, None)
+        if units_rule is not None:
+
+            sim = next((arg.sim for arg in args if isinstance(arg, SimArray)), None)
+            units_val = units_rule(*args, **kwargs)
 
             if isinstance(result, (tuple, list)):
-                units_val = (units_val,) if not isinstance(units_val, tuple) else units_val
+                if isinstance(units_val, tuple):
+                    if len(units_val) != len(result):
+                        raise ValueError("Output units must match result length")
+                else:
+                    units_val = itertools.cycle((units_val,))
                 result = type(result)(
                     _as_sim_array(r_i, u_i, sim) 
-                    for r_i,u_i in zip(result, itertools.cycle(units_val))
+                    for r_i,u_i in zip(result, units_val)
                     )
             else:
                 units_val = units_val[0] if isinstance(units_val, tuple) else units_val


apontzen:
Would it work instead to define a `@SimArray.ufunc_rule(np.gradient)` in the array module? This would be a more general solution and should still pass your test.

apontzen:
I think it would be fine to assume in the case that `result` is a tuple of `ndarray` each one can be converted to a `SimArray` and given the same units

pynbody/array/__init__.py
```diff
@@ -400,15 +400,27 @@ def __array_function__(self, func, types, args, kwargs):
...
+            for arg in args:
+                if isinstance(arg, SimArray):
+                    sim = arg.sim
+                    break
+            units_val = SimArray._ufunc_registry[func](*args, **kwargs)
+            
+            # If result is a tuple, convert each element
+            if isinstance(result, tuple):
+                result = tuple( r.view(SimArray) for r in result)
+                result = tuple(
```
apontzen, line 414: While this comprehension is quite clever, I think it's also a bit too cryptic. Also, it means duplicating essentially the same logic used for a single array below. I'd rather have something like

```python
def _as_sim_array(result, units, sim):
   result = result.view(SimArray)
   result.units = units
   result.sim = sim

...


if isinstance(result, tuple):
   result = tuple(_as_sim_array(r_i, units_val, sim) for r_i in result)
else:
   result = _as_sim_array(result, units_val, sim)
```

pynbody/array/__init__.py
```diff
@@ -966,6 +978,10 @@ def _comparison_units(ar, other):
 def _norm_units(a, *args, **kwargs):
     return a.units
 
+@SimArray.ufunc_rule(np.gradient)
+def _gradient_units(a, *varargs, **kwargs):
+    n = len(varargs)
```
apontzen, line 983: this `n` variable appears to be unused

wx-ys: I was considering cases where `varargs` might include a `SimArray` input, for example:
- ```np.gradient(SA1)/np.gradient(SA2)```
- ```np.gradient(SA1,SA2)```

I'm wondering if these two usages are equivalent, and whether `varargs` containing `SimArray` objects should affect the unit handling in `_gradient_units`.

apontzen:
Thanks - works well for the tested cases, but unfortunately I think there needs to be some more flexibility since we are trying to handle the general case. Consider this:

```python

v = SimArray([[1, 2, 3],
              [4, 5, 6]], 'K')
y = SimArray([1, 2], 'km')
t = SimArray([1, 2, 3], 's')
np.gradient(v, y, t)
```

I believe this should return

```python
(SimArray([[3., 3., 3.],
           [3., 3., 3.]], 'K km**-1'),
 SimArray([[1., 1., 1.],
           [1., 1., 1.]], 'K s**-1'))
```

but it currently returns

```python
(SimArray([[3., 3., 3.],
           [3., 3., 3.]], 'K km**-1'),
 SimArray([[1., 1., 1.],
           [1., 1., 1.]], 'K km**-1'))
```

i.e. it handles the differing units on the coordinate arrays incorrectly.

wx-ys:
I added a test case of this, but handling different types of `results` and `units` in `__array_function__` make this method a bit more complex.

apontzen:
Thanks!

Issue #943, opened 2025-10-22
Issues with the _override_units_system functionality


mtryan83:
**"Bug"**
The mechanism for overriding unit definitions: adding a `<filename>.units` file in the snapshot directory and calling `_override_units_system()` on the loaded snapshot (as done automatically when loading Tipsy and binary Gadget files) has a few snags:
1. The `<filename>.units` file needs to be the _exact_ filename, including the extension (so e.g. `snapshot_001.hdf5.units` ), which isn't documented anywhere, and the only way to know if it was used is by checking the subsequent units.
2. Blank lines are not permitted. This is fine (though I would argue overly strict and could be loosened), but the error produced is not helpful (`Unknown format for units file "<filename>.units"`). It would be helpful to provide either documentation on the format of the `.units` file and/or have a better error message (e.g. even just including which line caused the error).
3. **BUG**: If there is an error, from a blank line for example, the line that would _raise_ the `Unknown format...` error can produce a `TypeError` because `self.filename` is a PosixPath and isn't wrapped in `str()` like it is in the `try:` statement above. Note that this doesn't occur on the Gadget test data but does occur if attempting this on the test gizmo data.

**To Reproduce**
Preliminary: Add a blank line to the test_g2_snap.units file and copy the contents to `testdata/gizmo/snapshot_000.hdf5.units`.

```python
import pynbody

# To reproduce issue 2:
f = pynbody.load('testdata/gadget2/test_g2_snap')

# For issue 3
f = pynbody.load('testdata/gizmo/snapshot_000.hdf5')
f._override_units_system()
```

**Expected behaviour or result**
Expected no output or possibly debug output that the default unit system is being overwritten.

**Setup (please complete the following information):**
 - OS: ubuntu 24.04
 - Python version - 3.13
 - Numpy version - 2.3.3
 - Pynbody version - 2.3.1

You can access this information from python in the following one-liner:

```python
import sys, platform, pynbody, numpy; print("%s\r\n%s\r\npynbody %s\r\nnumpy %s\r\n"%(sys.version, platform.platform(), pynbody.__version__, numpy.__version__))
```

**Additional context**
This is related to issue #916 .

mtryan83:
> I'm not sure I agree the strictness of the path...

Sorry about that, I was in a hurry to write this and I think I came off rather harshly about all 3. They're all pretty minor quibbles that I mostly just wanted to have available as a reference/jumping off point for possible discussion/PRs. 

[...]

PR #945, opened 2025-10-23
Allow blank lines in .units file

mtryan83:
Added check to skip lines in the .units file that are whitespace only along with test. Improved error message to add line number that fails parsing.

Note this depends on #944 and fixes the second point in #943 .

### DIFF
--- a/tests/gizmo_test.py
+++ b/tests/gizmo_test.py
@@ -74,4 +74,3 @@ def test_allow_blank_lines():
     finally:
         # remove unitfile
         os.remove(unitfile_name)
-


tests/gizmo_test.py
```diff
@@ -17,3 +19,55 @@ def test_load_gizmo_file():
...
+        f.filename + ".units" 
+    # Note: Gizmo snapshots do not call _override_units_system by default
+    # so we call it manually
+    # This should produce an OSError because the file is formatted incorrectly
+    # not a TypeError because it's trying to concatenate a PosixPath and a str
+    with pytest.raises(OSError):
+        f._override_units_system()
+        
+    # remove unitfile
+    os.remove(unitfile_name)
```
apontzen: Would be good to put this in a `try`... `finally` to prevent any issues cleaning up

apontzen (commented):
Thanks! 

Small suggestion for cleaning things up here.

Also I don't think there's any reason why gizmo _doesn't_ call `_override_units_system`, so I'd favour putting it in there, rather than calling it manually in the test.

apontzen:
I think it’s just because it’s leftover functionality. It’s worth making its behaviour consistent across different files (or go the other way and strip it out entirely). If you are able to do so and document it, that’d be really great. But I agree it’s worth a separate PR.

I don’t understand the test failure either, but have seen similar things recently on interpolation routines that I haven’t managed to reproduce reliably. I think there might be some upstream numerical issue perhaps in scipy.interpolate. I’m trying rerunning the test.

PR #984, opened 2026-03-26
Fix bug in shared-memory mode loading DM-only tipsy snapshots.

### LINKED ISSUE #983: Shared memory mode fails with DM-only tipsy snapshots.
**Describe the bug**
When using pynbody together with tangos, if a simulation contains only DM particles then `tangos write` operations will fail when `--load-mode=server-shared-mem`.  This is because the `_load_main_file` method in `TipsySnap` creates arrays for gas and star particles even when there are only DM particles (e.g. "rho" and "temp" for gas).  Later in that method, when `_set_default_units` is called on these arrays, this will fail because the the array is a `SharedMemorySimArray`, and the object's `self.sim` attribute is `None`, causing a runtime error to be raised (`  RuntimeError("No link to SimSnap")`).

I believe the fix is to simply not create these arrays in `_load_main_file` unless the snapshot actually contains gas or stars.  I've created a branch on my fork that does this and seems to solve the problem, and I will put in a PR for it as soon as I've finished writing this bug report.

**To Reproduce**
To reproduce, build a tangos DB with a DM-only simulation and attempt to call `tangos write ` using the `server-shared-mem` load mode.  At least one calculation will fail, and the remaining "successful" ones may be incorrect.  I have been using "finder_mass" as the property

 - OS: Linux-5.14.0-427.35.1.el9_4.x86_64-x86_64-with-glibc2.34
 - Python version: 3.12.11 (main, Aug 14 2025, 00:00:00) [GCC 11.5.0 20240719 (Red Hat 11.5.0-11)]
 - Numpy version: numpy 2.4.1
 - Pynbody version: pynbody 2.3.2

bwkeller:
This PR prevents `_load_main_file` method of `TipsySnap` from creating arrays for gas or star particles if those families do not exist.  Prior to this, if pynbody is used to load a DM-only snapshot for tangos to write a property using `--load-mode=server-shared-mem`, the calculation will fail for the first halo calculation, and potentially silently write incorrect information for subsequent halos.

This fixes #983.

### DIFF
--- a/pynbody/snapshot/tipsy.py
+++ b/pynbody/snapshot/tipsy.py
@@ -171,6 +171,7 @@ class TipsySnap(SimSnap):
             for k in "rho", "temp", "metals":
                 if k in write:
                     self.gas[k].set_default_units(quiet=True)
+
         if 'star' in self.families():
             if "metals" not in list(self.star.keys()):
                 self.star._create_array("metals", zeros=False)
@@ -201,8 +202,6 @@ class TipsySnap(SimSnap):
         tbuf = bytearray(max_item_size * 10240)
 
         for fam, dtype in ((family.gas, self._g_dtype), (family.dm, self._d_dtype), (family.star, self._s_dtype)):
-            if fam not in self.families():
-                continue
             self_fam = self[fam]
             st_len = dtype.itemsize
             for readlen, buf_index, mem_index in self._load_control.iterate([fam], [fam], multiskip=True):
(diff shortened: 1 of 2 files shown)


apontzen:
Thanks @bwkeller. Is there a way to test specifically for the issue described in #983, to avoid regression?

bwkeller:
I think so.  I'll try and code one up as soon as I can and add it to the PR.

bwkeller:
I've added a test that checks whether this fix is working, the only test that has failed in the actions CI failed because of an HTTP 500 error downloading the `fg20` chemistry table.  If this looks good to you @apontzen I think it's ready to merge.

apontzen:
Thanks!

PR #987, opened 2026-04-10
Fix issue with loading larger grafic files

apontzen:
This related to an incorrect interaction between FortranFile and LoadControl

Gave rise to https://github.com/pynbody/genetIC/issues/130

Added a test


cphyc:
I think I'm a bit confused with why there was a bug in the first place. FortranFile reader _should_ read these ints properly, so this PR seems like it is just circumventing the underlying bug, isn't it?
I'm asking because the same code path is used for reading RAMSES outputs if I remember correctly, so it might also affect it.

apontzen:
No, the issue is more subtle than that. It’s because partial loading may break up a single Fortran record into multiple load blocks. (See the added test for contrived but informative examples.)

cphyc:
Ah, I see! Sorry for the noise then

Issue #994, opened 2026-06-22
Test failure due to healpy calling a deprecated function


jchelly:
**Describe the bug**

The test `tests/sph_image_test.py::test_render_stars_spherical` raises a PendingDeprecationWarning, which is converted into an error by pynbody's pytest configuration.  I think this is due to healpy calling a deprecated function in matplotlib.

**To Reproduce**

With pynbody installed from source in editable mode:
```
pytest ./tests/sph_image_test.py
```
which results in
```
FAILED tests/sph_image_test.py::test_render_stars_spherical - PendingDeprecationWarning: The set_over function will be deprecated in a future version. Use cmap.with_extremes(over=...) or Colormap(over=...) instead.
```
**Expected behaviour or result**

The test should pass!

**Setup (please complete the following information):**
 - OS
 - Python version
 - Numpy version
 - Pynbody version

3.12.4 (main, Jul  4 2024, 14:42:26) [GCC 11.4.1 20231218 (Red Hat 11.4.1-3)]
Linux-5.14.0-611.55.1.el9_7.0.3.x86_64-x86_64-with-glibc2.34
pynbody 2.4.2
numpy 2.5.0

**Additional context**

This is with healpy 1.19.0. Maybe this particular warning should be added to the ignore list in pynbody's pyproject.toml until it's fixed in healpy.

apontzen:
Thanks, fixed in #995

PR #995, opened 2026-06-22
Filter out deprecation warnings from healpy

jchelly:
When running tests with pytest, this filters out PendingDeprecationWarnings generated by healpy calling deprecated matplotlib functions. Addresses #994.

There's another test failure (on this branch and on master) in hmf_test.py, which I think is not related to this one.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -17,12 +17,12 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-2025]
         python-version: ["3.11", "3.12", "3.13", "3.14"]
-        numpy-version: ["2.0.2", "2.3.3"]
+        numpy-version: ["2.1.0", "2.5.0"]
         exclude:
           - python-version: "3.14"
-            numpy-version: "2.0.2"
-          - python-version: "3.13"
-            numpy-version: "2.0.2"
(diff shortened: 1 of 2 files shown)


jchelly:
I think the tests might have passed here because github is testing against a few specific numpy versions. On Cosma I get a failure in hmf_test.py that seems to be caused by https://github.com/cmbant/CAMB/issues/202 and https://github.com/scipy/scipy/issues/25471 .

apontzen:
The intent is to test against the newest and oldest supported libraries in SPEC0 but pin versions to disentangle library updates from feature changes.

If you're able to update the bracketing library versions in this PR please do!

jchelly:
The hmf_test failure happens because of a bug introduced in scipy 1.18.0. I've set "scipy>=1.12.0,<1.18.0" in pyproject.toml for now, but we could change to "!=1.18.0" if the next release is fixed.

The oldest supported numpy in SPEC0 is now 2.1 and the oldest python is 3.12. I've updated the versions tested in the github workflow.

jchelly:
I'm not having much luck with the readthedocs build. Downloading the test data failed with a 502 bad gateway on the first attempt then a 403 forbidden on the second.

apontzen:
Thanks, some sort of network issue presumably. I've tried re-triggering the build.

PR #997, opened 2026-07-13
Several fixes for downloading test data on github

jchelly:
This PR makes several changes intended to reduce failures when running the pynbody tests through github:

  * osfclient file objects are cached to reduce the number of requests. This avoids downloading the directory listing again for each file so fetching the test data takes ~20 requests instead of ~100.
  * The .npz files used in ionfrac_test.py are downloaded and cached in the same way as the test data, so we don't have all jobs in the test matrix requesting them in parallel.
  * The fetch_testdata workflow should no longer save a corrupted cache if downloads fail (see issue #968)

OSF unauthenticated requests seem to be limited to 100 per hour. I think one test run will now be well below this limit and we shouldn't generate any requests once the test data and tables are cached.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -9,9 +9,11 @@ defaults:
 jobs:
   fetch-testdata:
     uses: ./.github/workflows/fetch-testdata.yaml
+  fetch-tables:
+    uses: ./.github/workflows/fetch-tables.yaml
 
   build:
-    needs: fetch-testdata
+    needs: [fetch-testdata, fetch-tables]
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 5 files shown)


apontzen:
Thanks John. 

This is a bit fussy of me but I would prefer if possible not to hard-code the osf archive ID into the workflow yaml. It could lead to unexpected test failures if we change the tables later and forget to update the yaml. In future we might use a different storage location for these tables (e.g. just a plain s3 bucket somewhere). 

Ideally, we would expose a way to get these tables using the existing method `_download_ionfracs` in `ionfrac.py`, so that if this method is changed then the cache also automatically points to the right place.

Other than this, everything looks good to me.

jchelly:
Here's a version where the same function is used for pre-caching the tables and downloading them when they're needed. I put the function in test_utils because it needs get_osf_file_object() and imports are tricky when pynbody isn't installed yet.

apontzen:
I am probably being really paranoid but I foresee a potential failure down the line if we edit `test_utils` without realising that `ionfrac` itself depends on them. In particular, since the pre-caching would already have downloaded the ion tables, this would not be caught by CI.

However, I can see the problem with using an implementation inside ionfrac without installing pynbody itself. 

I would therefore suggest compromising on two simple changes, first to make `_download_ionfrac_table` public (i.e. remove the leading `_`) and second to add a docstring highlighting this is actually used by ionfrac itself. Hopefully these two hints will help guard against a future error.

apontzen:
Thanks!

PR #998, opened 2026-07-14
Fix for occasional bisect() failures in calculating virial radii

bwkeller:
The initial guess for r_max in virial_radius currently only uses the x positions, even though the comment says it uses all positions.  I've updated this to use x,y,z positions by taking the max/min of pos.  This should guarantee convergence of the later bisect call even for pathological cases with very aspherical halos.

### DIFF
--- a/pynbody/analysis/halo.py
+++ b/pynbody/analysis/halo.py
@@ -145,7 +145,7 @@ def virial_radius(sim, cen=None, overden=178, r_max=None, rho_def='matter'):
     """
 
     if r_max is None:
-        r_max = (sim["pos"].max() - sim["pos"].min())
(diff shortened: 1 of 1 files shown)


apontzen:
I think the point is right but the solution is possibly wrong... if, say, the halo is centred at (1000,1,1) and has ellipsoidal radius (2,1,1), this would give the initial guess as 1000 not 2. Will cause considerable slowdown.

My (untested) suggestion would be something more like

```
(f['pos'].max(axis=0) - f['pos'].min(axis=0)).max()
```
There may be some even more efficient way to express this...

bwkeller:
Good point!  I've replaced that line with a call to numpy's `ptp` method, which I think is the fastest way to calculate the range along a given axis.  Has the additional bonus of being correct :P

PR #1003, opened 2026-07-21
Implement partial loading of Swift snapshots using the GadgetHdfSnap take parameter

jchelly:
Currently the SwiftSnap snapshot class implements cutting out regions by making a temporary HDF5 file with virtual datasets which refer to the selected parts of the snapshot. This PR replaces it with an implementation based on the partial access mechanism already present in the parent GadgetHdfSnap class. The approach is to treat the `take_region` and `take_cells` parameters as alternative ways to specify the array of particle indexes which would otherwise be supplied by the `take` parameter.

The motivation for this is to make it easier to allow remote access to Swift snapshots using the hdfstream module. When the snapshot files are accessed using http we can't create a local virtual dataset that refers to them.

I'm marking this PR as a draft for now because it needs some additional tests. E.g.

  * Selecting a region from a multi-file snapshot with multiple particle types
  * Reading a multi-file snapshot where some files are missing particles of some type
  * Reading a snapshot where some particle type is present in the cell metadata but not this particular snapshot
  
Test cases can probably be set up by using the single file DM+gas snapshot from the test data as a template if we don't want to add more files to the test data.

### DIFF
--- a/pynbody/snapshot/swift.py
+++ b/pynbody/snapshot/swift.py
@@ -143,6 +143,17 @@ class SwiftMultiFileManager(_GadgetHdfMultiFileManager):
     def _all_group_names(self):
         return self[0]['Cells/Counts'].keys()
 
+    def iter_particle_groups_with_name(self, hdf_family_name):
+        """
+        Swift snapshots can contain zero length datasets, which we need to skip.
+        TODO: should this be handled in GadgetHdfSnap somewhere?
+        """
+        for hdf in self:
+            if hdf_family_name in hdf:
+                if self._size_from_hdf5_key in hdf[hdf_family_name]:
+                    n = hdf[hdf_family_name][self._size_from_hdf5_key].shape[0]
+                    if n > 0:
+                        yield hdf[hdf_family_name]
 
 class ExtractScalarWrapper:
     def __init__(self, underlying):
--- a/pynbody/snapshot/swift.py
+++ b/pynbody/snapshot/swift.py
@@ -13,39 +13,6 @@ from .gadgethdf import GadgetHDFSnap, _GadgetHdfMultiFileManager
 
 class SwiftMultiFileManager(_GadgetHdfMultiFileManager):
 
-    def __init__(self, filename: pathlib.Path, take_cells, take_region, mode='r'):
-        self._take_cells = take_cells
-
-        if take_cells is not None and take_region is not None:
-            raise ValueError("Either take_cells or take_region must be specified, not both")
-
-        if take_cells is not None or take_region is not None:
-            # we need to avoid loading a VDS file, as it won't have the info needed to make our own VDS
-            # pointers to the right cells
-            try:
-                with h5py.File(filename, mode='r') as f:
-                    if f['Header'].attrs['Virtual'][0] == 1 and filename.suffix == '.hdf5':
-                        # if we simply strip off .hdf5, GadgetHDFSnap will find the underlying files (hopefully..)
-                        filename = filename.with_suffix('')
-                    # it's not a virtual file anyway, so we're ok
-            except FileNotFoundError:
-                pass # perfect, this is either going to be pieced together by pynbody, or it'll fail anyway
-        super().__init__(filename, mode)
-
-        if take_region is not None:
-            # convert take to take_cells
-            take_cells = self._identify_cells_to_take(take_region)
-
-        # hopefully the shenanigans above mean we don't end up with a VDS, but just in case we do, check again:
-        if self.is_virtual() and take_cells is not None:
-            raise ValueError("Can't take a subset of cells from a VDS-based HDF5 file pointing to multiple underlying files")
-
-        self._make_hdf_vfile(take_cells)
-
-    def _identify_cells_to_take(self, take):
-        centres = self[0]['Cells/Centres'][:]
-        return np.where(take.cubic_cell_intersection(centres))[0]
-
     def get_unit_attrs(self):
         return self[0].parent['InternalCodeUnits'].attrs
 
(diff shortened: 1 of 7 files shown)


jchelly:
This now appears to be working correctly but I suspect I might have reintroduced the problem with Arepo tracer particles mentioned in #907 by allowing zero length groups.

jchelly:
If I download an Illustris snapshot file I can see that it has a PartType3 group with ParentID and TracerID only (no ParticleIDs). Loading the snapshot with pynbody works, but it treats the tracers as DM particles in the `_family_to_group_map`:
```
In [9]: snap._family_to_group_map
Out[9]: 
{<Family dm>: ['PartType1', 'PartType3'],
 <Family star>: ['PartType4'],
 <Family gas>: ['PartType0']}
```
I think the fix here is that groups which do not contain the key `_size_from_hdf5_key` should not be considered when determining what families we have. Then we can read PartType3 in zoom simulations where it contains dark matter and skip it for runs where it contains tracers.

jchelly:
Pytest's tmp_path fixture leaves test data lying around under /tmp/, which generated a complaint from cosma support! The generated test snapshots could be stored somewhere under testdata and reused. That would be more convenient for debugging test failures too.

tests/swift_test.py
```diff
@@ -281,3 +282,139 @@ def test_planetary_physical_units():
...
+
+    # But we should still always have gas and dm families with arrays
+    for name in ("pos", "vel", "mass", "iord", "grp"):
+        assert name in f.dm.loadable_keys()
+        assert name in f.gas.loadable_keys()
+
+    # We should be able to read the gas coordinates, and get an empty array
+    # if the selection does not cover the region with gas.
+    gas_pos = f.gas["pos"]
+    assert gas_pos.shape == (274,3) if expect_gas else (0,3) # cell 337 has 274 gas particles
```
apontzen: I think the intent here is probably `gas_pos.shape == ((274,3) if expect_gas else (0,3))` but without the brackets I don't think it'll parse that way

pynbody/snapshot/swift.py
```diff
@@ -13,38 +7,120 @@
...
+                cell_file_offset = self._cells[name]["offsets"]
+                particles_per_cell = self._cells[name]["counts"].copy()
+                # Get the original number of files in the full snapshot
+                nr_files_total = np.amax(cell_file_index) + 1
+                # Treat cells in files we don't open as empty, because we're
+                # computing an index into the subset of files we opened.
+                particles_per_cell[self._file_mask[cell_file_index]==False] = 0
+                # Compute the offset to the first particle in each file in the subset.
+                # Note that the file index refers to the full set of snapshot files here.
+                particles_per_file = np.bincount(cell_file_index, weights=particles_per_cell, minlength=nr_files_total)
```
apontzen: I think bincount returns float64 when called with weights, even if the weights are themselves integers. Probably this doesn't matter but to make intent clear perhaps we want an `.astype(np.int64)` or something here.

pynbody/snapshot/swift.py
```diff
@@ -13,38 +7,120 @@
...
-                with h5py.File(filename, mode='r') as f:
-                    if f['Header'].attrs['Virtual'][0] == 1 and filename.suffix == '.hdf5':
-                        # if we simply strip off .hdf5, GadgetHDFSnap will find the underlying files (hopefully..)
-                        filename = filename.with_suffix('')
-                    # it's not a virtual file anyway, so we're ok
-            except FileNotFoundError:
-                pass # perfect, this is either going to be pieced together by pynbody, or it'll fail anyway
-        super().__init__(filename, mode)
-
+    def __init__(self, filename, mode='r', take_swift_cells=None, take_region=None):
```
apontzen, line 13: I'm concerned this no longer calls `super().__init__` and instead reimplements the file discovery.

I can see why this has happened, but I worry it could cause confusion down the line if they drift apart somehow. Could we factor `_GadgetHdfMultiFileManager.__init__` into something like `_discover_filenames` and `_select_files`, and override appropriately here?

pynbody/snapshot/swift.py
```diff
@@ -265,3 +263,23 @@ def halos(self, **kwargs):
...
+        from the group's attributes, then all of the keys must be datasets.
+        """
+        if "NumberOfFields" in group.attrs and group.attrs["NumberOfFields"][0] == len(group):
+            return list(group.keys())
+        else:
+            keys = []
+            def _append_if_array(to_list, name, obj):
+                if not hasattr(obj, 'keys'):
+                    to_list.append(name)
+            group.visititems(functools.partial(_append_if_array, keys))
```
apontzen: according to pycharm, functools is not actually imported so I can't work out how this code works. I think possibly the missing NumberOfFields case is not getting tested anywhere, and would crash if tried?

pynbody/snapshot/swift.py
```diff
@@ -13,38 +7,120 @@
...
+
+        # Compute particle "take" indices corresponding to the selected cells
+        take = []
+        ptype_offset = 0
+        for family in families_ordered:
+            for name in family_to_group_map[family]:
+                # Find the cell metadata
+                cell_file_index = self._cells[name]["files"]
+                cell_file_offset = self._cells[name]["offsets"]
+                particles_per_cell = self._cells[name]["counts"].copy()
```
apontzen, line 89: This may be baseless but I worry about what happens if the metadata is in some way corrupt. Could we add an assert that at least checks that `particles_per_cell.sum()` agrees with the total number of particles expected, or something like that? This might catch the worst problems.

pynbody/test_utils/split_swift_snapshot.py
```diff
@@ -0,0 +1,262 @@
...
+    return m.hexdigest()
+
+
+@contextmanager
+def staging_dir(final_path):
+    final_path = Path(final_path)
+    final_path.parent.mkdir(exist_ok=True)
+    with TemporaryDirectory(dir=final_path.parent, prefix=f".{final_path.name}.") as tmp:
+        tmp_path = Path(tmp)
+        tmp_path.mkdir(exist_ok=True)
```
apontzen: I think TemporaryDirectory already will have created this so it's redundant

apontzen (changes requested):
Thanks @jchelly! I have some small inline comments. 

I'm also unclear whether this reintroduces the problem @wx-ys was guarding against in #907. That check dropped particle groups that measured as zero length, with the comment:

> "This is important for some formats like Arepo where tracer particles might be defined but not present, which can cause issues with HaloCatalogue."

Two separate questions, I think. First, removing `_remove_empty_particle_groups` means zero-length groups are now retained — I wonder if @wx-ys is able to comment on whether this may be a problem in light of the comment above.

Second, the new criterion at gadgethdf.py:662 silently drops any PartTypeN lacking ParticleIDs, taking the whole family with it where that's the only group for the family. @jchelly's diagnosis of the tracer case is convincing and I think this is the more correct criterion, but silently losing a family might be a worse failure mode than the misclassification it fixes... should we at least warn when a group is skipped this way?

I've become hazy about what the right behaviour is, not helped by the fact that we have no test covering any of it — `_remove_empty_particle_groups` went in without one, and it's been released since v2.3.0. Thoughts most welcome.

The other matters I've noted in the comments are very small / much simpler.

Thanks!

wx-ys:
> If I download an Illustris snapshot file I can see that it has a PartType3 group with ParentID and TracerID only (no ParticleIDs). Loading the snapshot with pynbody works, but it treats the tracers as DM particles in the `_family_to_group_map`:
> 
> ```
> In [9]: snap._family_to_group_map
> Out[9]: 
> {<Family dm>: ['PartType1', 'PartType3'],
>  <Family star>: ['PartType4'],
>  <Family gas>: ['PartType0']}
> ```
> 
> I think the fix here is that groups which do not contain the key `_size_from_hdf5_key` should not be considered when determining what families we have. Then we can read PartType3 in zoom simulations where it contains dark matter and skip it for runs where it contains tracers.

I think this is a better approach. The reason ``_remove_empty_particle_groups `` was introduced earlier is explained here:
https://github.com/pynbody/pynbody/pull/907#discussion_r2175153208

tests/swift_test.py
jchelly: Thanks, that's a good catch. I'm asserting either True or (0,3) depending on the test case so the assert always passes.

pynbody/snapshot/swift.py
jchelly: Yes, I think I have float64s propagating to quantities I intended to be integers. Bincount should be ok up to about 2^53 if we convert the result, but maybe its better to use np.add.at instead to keep everything in integer types?

pynbody/test_utils/split_swift_snapshot.py
jchelly: Agreed. I'll remove the call.

pynbody/snapshot/swift.py
apontzen: Yes I wasn't aware of that possibility. It sounds more idiomatic, even if in reality it can't matter very much.

pynbody/snapshot/swift.py
jchelly: I wanted the code to fall back to the Gadget HDF5 behaviour if non-dataset objects were added to Swift PartType groups in the future, but there isn't a test case for that so I didn't catch the missing import.

I'm now not sure what's best here. Matthieu says that PartType groups in Swift snapshots only ever contain datasets, in which case we could return the list of keys without any checks. Maybe we ought to treat NumberOfFields as a consistency check on the snapshot and raise an exception if it doesn't match the number of keys?

apontzen: That sounds reasonable - fail with a clear message

pynbody/snapshot/swift.py
jchelly: I've added some checks. It looks like the planetary example doesn't have valid cells because the counts are all zero. But we can skip reading the cells at all if we're not reading a region.

pynbody/snapshot/swift.py
jchelly: I was trying to avoid reopening file 0 repeatedly, although I'm not sure how much that matters. I've changed it so that if `_GadgetHdfMultiFileManager.__init__` opens a file it gets stored in `self._open_files`. Then `SwiftMultiFileManager.__init__` can call `super().__init__` and after that it should be ok to open the file with self[0]. Pruning the file list can be done after that.

apontzen:
Thanks! A couple of final requests...

## Gracefully dealing with NumberOfFields

These changes are looking good but the raise in `_get_hdf_allarray_keys` is a problem. For example,

```
f = pynbody.load('testdata/SWIFT/snap_0150.hdf5')
f['test_array'] = np.random.uniform(0, 1, len(f))
f['test_array'].write()
pynbody.load('testdata/SWIFT/snap_0150.hdf5')   # ValueError — snapshot now permanently unloadable
```

I think the fix is to warn and fall back to `GadgetHDFSnap._get_hdf_allarray_keys(group)` when `NumberOfFields` is inaccurate. 

(Separately, `write_array` arguably ought to bump NumberOfFields for SWIFT. )

## Good errors for invalid metadata

It's interesting that the planetary snapshot just has zeros for its cell counts. I think in this case there should be a special exception, i.e. rather than the generic-sounding `The total number of particles in all cells does not match the snapshot header`, perhaps when the counts are all zero it could be more like `No spatial index found in this snapshot; unable to load a region.`

jchelly:
The planetary snapshot only has the first 1000 particles from a larger snapshot. `NumPart_Total` in the header is much larger than 1000. Maybe the cell metadata was cleared because it would have been wrong? I don't think Swift would write a snapshot with all cell counts zero because it uses the cells itself.

There are some more issues related to writing:
  * I'm not sure what happens (or should happen) if we try to write after we read a region or cells. The `write_array()` implementation in gadgethdf.py looks like it wont work in that case, so maybe we just raise a NotImplementedError for now.
  * I've forgotten to pass the mode parameter from `SwiftSnap` to the parent `GadgetHdfSnap` constructor
  * If I load an example Swift snapshot with mode="r" and try to write to it in an interactive session, reopening in writeable mode fails because the file is already open. Somehow it works when pytest does the same thing, so that needs some investigation.

jchelly:
Ok, so GadgetHdfSnap doesn't take a mode parameter either, and I think the reopening issue was due to having a left over reference to the snapshot in my interactive session. I've added a NotImplementedError on writing to a region.

jchelly:
This now updates NumberOfFields on writing an array (by assuming that all keys in Swift particle groups are datasets), and I've added a test to check that zero sized datasets don't cause problems for write_array().

apontzen (approved):
I think this looks good now, shall I go ahead and merge?

wx-ys (approved):

jchelly:
Yes, thanks, I'm happy for this to be merged now.

PR #1008, opened 2026-08-01
Enhance partial loading performance with HDF5 chunk cache configuration

apontzen:
I am seeing very poor performance when loading part of a SWIFT file. This PR addresses the issue.

- Added configuration options for HDF5 chunk cache size and slots in default_config.ini.
- Implemented a custom _open_hdf_file function to optimize file opening with chunk caching.
- Updated SwiftMultiFileManager to utilize the new chunk cache settings for efficient partial reads.
- Introduced a coalesce_cells_into_slices method to minimize virtual mappings for contiguous cells.
- Added tests to verify the effectiveness of the new optimizations.

@jchelly I am not sure how this interacts with your #1003.  Likely it removes the need for some of these optimisations - but perhaps not all. Your comments would be very welcome!

### DIFF
--- a/pynbody/default_config.ini
+++ b/pynbody/default_config.ini
@@ -109,6 +109,19 @@ StellarFormationTime: tform
 BHFormationTime: tform
 Potential: phi
 
+[gadgethdf]
+# Size of the HDF5 chunk cache, in bytes, used when opening gadget-like HDF5 files (including swift).
+# Simulation snapshots are often written as large compressed chunks, and HDF5 must decompress an entire
+# chunk to satisfy even a small read. If the cache cannot hold a chunk, that decompression is repeated
+# for every read, which makes partial loading extremely slow. The default of 64MB is enough to hold a
+# few chunks of a typical snapshot. Note that this is a cap rather than an allocation, and that whole-
+# chunk reads bypass the cache, so full snapshot loads are unaffected by this setting.
+chunk-cache-nbytes: 67108864
+
+# Number of slots in the HDF5 chunk cache. Should be a prime number, ideally at least ten times the
+# number of chunks that can fit in the cache.
+chunk-cache-nslots: 5051
+
 [gadgethdf-type-mapping]
 # GadgetHDF stores six different particle types (numbered 0 to 5). This specifies how they map
 # onto pynbody particle types by default. To override, you can either make your own configuration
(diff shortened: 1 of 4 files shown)


jchelly:
The chunk cache optimization will still be useful for #1003. The coalesce_cells_into_slices() method shouldn't be necessary because merging of reads for adjacent cells will be dealt with by the take machinery. It might be nice to allow the take parameter to be a list of slices so we don't convert to a full array of indexes and back again.

When reading from hdf5 files we could select multiple slices with H5Sselect_hyperslab() and then read them with a single H5Dread() call. I investigated this for swiftsimio (https://github.com/SWIFTSIM/swiftsimio/pull/311) and found that it made the test suite quite a bit faster. It seemed like each h5py read_direct() had a bit of overhead which you don't get with the low level API, but I don't know how much it matters for larger reads.

In #1003 I also found that fetching the dataset names could be a bottleneck in flamingo and colibre where there can be ~60 datasets in a group. Getting just the names is fast but opening them to see if they're groups or datasets is much slower if the snapshot hasn't been read recently. Swift PartType groups have a NumberOfFields attribute so I've been assuming that if that exists and equals the number of keys then all of the keys are datasets.

apontzen:
Many thanks @jchelly. In this case it would be good to get #1003 merged soon - currently it's still draft, and I see your comments about what remains to be done but I'm not sure how much work it will be in practice?

jchelly:
I've written the extra tests I wanted in #1003 and they pass now, so we could take a look at merging it.

apontzen:
Closing this: one half has landed elsewhere, and the other half is no longer needed.

**The chunk-cache configuration is carried by #1009.** `default_config.ini` gains `chunk-cache-nbytes`/`chunk-cache-nslots`, and `_open_hdf_file` is used for every real file the gadget-HDF multi-file manager opens (`pynbody/snapshot/gadgethdf.py:106` and `:133`). SWIFT inherits that manager and, since #1003, opens no HDF5 files of its own, so SWIFT partial loads get the cache — which was the problem that prompted this PR.

**`_coalesce_cells_into_slices` is obsolete, as @jchelly predicted here.** It merged runs of adjacent cells into virtual-dataset mappings; #1003 removed the virtual-dataset path from the SWIFT loader altogether, in favour of feeding the generic `take` machinery. There is no `_hdf_vfile` and no per-cell mapping construction left for it to optimise, and its two tests inspected structures which no longer exist.

The remaining two suggestions from the review above — allowing `take` to be a list of slices instead of a per-particle index array, and using `H5Sselect_hyperslab` with a single `H5Dread` for multi-slice reads — are not implemented, and are now tracked in #1024 so they are not lost. The third, the `NumberOfFields` shortcut, was adopted in #1003.

---

PR #1009, opened 2026-08-02
Add support for incomplete halos when partial loading

apontzen:
Partial loading and halo catalogues have never fitted together. A halo finder assigns particles to a halo by ID or by file position, and if some of those particles were never read from disk, the catalogue's membership information no longer describes anything the snapshot can supply. Until now that produced either an `IllegalIordError` from deep inside the iord mapping machinery, or silently truncated halos.

This PR makes such halos a first-class concept: a halo which the finder assigns particles that are not present in the snapshot is *incomplete*, accessing it raises a clear error, and the catalogue can say in advance which halos are affected. HBT+ is wired up as the first finder to use it; #1011 builds on this API to cover the rest.

## The user-facing API

```python
f = pynbody.load("snap", take=slice(1000))
h = f.halos()

h.keys()             # every halo in the catalogue, whether or not it can be loaded
h.complete_keys()    # the subset which can actually be retrieved
h.is_complete(17)    # ... for one halo
h.get_complete_mask()  # ... in halo *index* order, so it can filter get_properties_all_halos()
h[17]                # raises IncompleteHaloError if particles are missing
```

`IncompleteHaloError` carries `halo_number` and `num_missing_particles`, and its message points at `complete_keys()`. Halo *properties* remain available for incomplete halos; it is only particle access which fails.

`keys()` now returns a read-only numpy array rather than whatever the number mapper happened to hold, since it may be a view of the catalogue's internal numbering. Establishing completeness needs the particle lists for all halos, so these methods call `load_all()`; pass `load_all_if_required=False` to get a `RuntimeError` instead, for catalogues too large to load eagerly.

## How a finder opts in

`HaloCatalogue` gains two helpers wrapping the iord mapper, so finders no longer touch `_iord_to_fpos` directly:

- `_map_iords_to_fpos(iords) -> (fpos, num_missing)`, for `_get_all_particle_indices`: missing particles are dropped and counted, so loading every halo does not fail on account of one bad halo.
- `_map_iords_to_fpos_one_halo(iords, halo_number)`, for `_get_particle_indices_one_halo`: raises `IncompleteHaloError` directly.

The counts go to `HaloParticleIndices` as `num_missing_particles`, which is all that is needed — `complete_keys()` and friends are then derived automatically. Catalogues which delegate `load_all()` to another catalogue rather than populating their own index lists must instead provide `_is_loaded()` and `_get_complete_mask()`; `SubhaloCatalogue` and `HBTPlusCatalogueWithGroups` do so, mapping the underlying answer into their own halo indexing.

## Supporting changes

**Iord mapping** (`halo/details/iord_mapping.py`). `map_ignoring_order` gains `allow_missing`, returning the `NO_OFFSET` sentinel instead of raising, and the previously bare `ValueError` becomes a named `IllegalIordError` whose message names the offending IDs. All three mapper implementations honour it, including `IordOffsetModifier`, which must not shift the sentinel along with the genuine offsets. The dense mapper also no longer lets an out-of-range or negative iord wrap around the end of its lookup table — a latent correctness bug, since numpy would silently return an unrelated particle. The usual case (nothing missing) still costs a single reduction and one fancy-index, with no temporaries.

**`take` validation** (`chunk/__init__.py`). Partial loading only works if the requested ids ascend, because the readers walk the file forwards; previously tipsy quietly coped, gadgethdf failed with an `IndexError` from inside its chunk handling, and repeated ids produced a snapshot containing the same particle twice — which breaks the one-to-one iord mapping the halo machinery now relies on. Both are rejected up front, where they can be explained. `take` also accepts a `slice`, normalised against the on-disk particle count so that everyday forms like `slice(1000)` and `slice(-100, None)` mean what they look like.

**HDF5 chunk cache** (`default_config.ini`, `gadgethdf.py`). Snapshots are routinely written as compressed chunks of megabytes, and with the default 1MB cache HDF5 re-decompresses an entire chunk for *every* partial read, making partial loading orders of magnitude slower than reading the whole file. Gadget-like files are now opened with a 64MB cache, configurable via `chunk-cache-nbytes`/`chunk-cache-nslots`. It is a cap rather than an allocation, and whole-chunk reads bypass the cache, so full loads are unaffected.

**`uint64` iords** (`util/_util.pyx`). The fused integer types used by `binary_search` and friends did not include `uint64`, which is what SWIFT writes, so the sparse mapper — precisely the one selected in the partially loaded case — raised `TypeError: Function call with ambiguous argument types`.

**Docs.** The partial-loading material in `docs/loaders.rst` was a duplicate of the treatment in `docs/tutorials/parallelism.rst`; it is removed, and the tutorial gains sections on `take`/`take_region` and on incomplete halos. `pynbody.load`'s docstring documents `take` and `take_region`, which it previously did not.

## Tests

`tests/halos_test.py` (+395) covers the API against synthetic catalogues, including the cases most likely to break: halo numbers not in index order (`NonMonotonicHaloNumberMapper`), a catalogue where nothing is complete, a delegating catalogue which forgets to provide the hook, mask shape/dtype validation, read-only-ness and caching of both `keys()` and the mask, and the equivalence of the answer with and without `load_all()`. `tests/hbtplus_test.py` (+124) exercises real partially loaded HBT+ data across `None`/`length-order`/`track` numbering and the combined group catalogue. `tests/partial_tipsy_test.py` (+46) pins the new `take` validation and slice handling.

## Notes for review

- **Version target.** 2.6.0 shipped while this branch was open, so the new API is marked `.. versionadded:: 2.7.0` and `__version__` moves to 2.7.0 (master was left at 2.6.0 after the release). The 2.6.0 markers elsewhere in the tree describe what genuinely shipped and are untouched.
- **`keys()` is now read-only.** Downstream code which mutates the returned array will break. This seems worth it — handing out a writable view of the catalogue's own numbering was a bug waiting to happen — but it deserves a release note.
- **Only HBT+ reports completeness in this PR.** Every other finder still behaves as before, which for the ID-based finders means `IllegalIordError`, and for the file-position finders means silently wrong particles. #1011 addresses both.


pynbody/chunk/__init__.py
```diff
@@ -83,9 +83,30 @@ def __init__(self, family_slice: dict[family.Family, slice], max_chunk: int, cla
...
-            self._ids = np.asarray(clauses, dtype=np.int64) 
+        if isinstance(clauses, slice):
+            # A slice indexes the particles as they lie on disk, so normalise it against that count before
+            # expanding it. Without this, the omitted start/step of an everyday slice such as slice(10) reach
+            # np.arange as None, and negative bounds like slice(-100, None) are taken literally.
+            disk_num_particles = self._disk_family_slice[self._ordered_families[-1]].stop
+            start, stop, step = clauses.indices(disk_num_particles)
+            if step < 0:
+                raise ValueError("A slice passed as 'take' must select particles in the order they appear on "
+                                 "disk, so a negative step is not supported. Reverse the snapshot after "
```
apontzen: Not clear whether it will be possible to 'reverse the snapshot after loading it' (e.g. family order can't be reversed) so don't give this advice. Just give the factual first sentence.

pynbody/halo/details/iord_mapping.py
```diff
@@ -7,13 +7,58 @@
...
+
 
 class IordToOffset(abc.ABC):
     @abc.abstractmethod
-    def map_ignoring_order(self, i: np.ndarray | int) -> np.ndarray | int:
+    def map_ignoring_order(self, i: np.ndarray | int, allow_missing: bool = False) -> np.ndarray | int:
         """Given an array of iord values, return the corresponding fpos values.
 
-        Warning: The returned values are not guaranteed to be in the same order as the input iord array."""
+        Warning: The returned values are not guaranteed to be in the same order as the input iord array.
```
apontzen: change this to a proper RST warning block

pynbody/halo/__init__.py
```diff
@@ -306,9 +318,115 @@ def __iter__(self) -> Iterable[Halo]:
...
+
+        .. versionadded:: 2.7.0
+
+        """
+        all_numbers = self.keys()
+        complete_mask = self.get_complete_mask(load_all_if_required)
+
+        # NB the halo numbers are not necessarily in halo index order (see NonMonotonicHaloNumberMapper),
+        # so the mask must be explicitly mapped rather than applied directly
+        all_indices = np.asarray(self.number_mapper.number_to_index(all_numbers), dtype=np.intp)
```
apontzen: is this cast needed? I'd have thought number_to_index already returns a positive integer numpy array so I am not sure what this is doing here

apontzen: You're right, it isn't needed here — removed in 0c4c876.

All three built-in mappers already return an integer array for an array argument: `MonotonicHaloNumberMapper` returns `np.searchsorted`'s output, `NonMonotonicHaloNumberMapper` indexes an `argsort` result with it, and `SimpleHaloNumberMapper` subtracts an offset from the integer numbers it generated itself. The cast could only ever have mattered for a third-party mapper subclass, and it would have failed loudly at the indexing step anyway.

The similar-looking cast in `SubhaloCatalogue._get_complete_mask` is a genuinely different case and stays: the halo numbers there come from the caller rather than from the mapper, and are an empty float array when a group has no subhalos — which is what `test_subhalo_catalogue_completeness_with_no_subhalos` exercises. I've left the comment there recording why.

---

PR #1010, opened 2026-08-09
Allow halo catalogues to be expressed as arrays and rebuilt elsewhere

apontzen:
Sharing a halo catalogue between processes currently requires the consumer to reach into pynbody's internals. [tangos does exactly that](https://github.com/pynbody/tangos/blob/master/tangos/parallel_tasks/pynbody_server/shared_object_catalogue.py), reassembling the catalogue, its number mapper and its particle index lists by hand. That ties external code to internals which are free to change, and silently loses anything it does not explicitly handle — the completeness information from #1009 and #1011 being a case in point.

This gives that job an API instead.

## The API

```python
state = h.get_portable_state()          # dict of numpy arrays and python primitives
...                                     # move it: shared memory, pickle, a file
h = pynbody.halo.HaloCatalogue.from_portable_state(state, f)   # a live catalogue on f
```

The consumer needs no knowledge of what any individual array is for. `portable.map_arrays(state, mapper)` walks a state and replaces every array in it, so a task manager can swap them all for shared-memory references without understanding the structure. The worked example lives at the bottom of the parallelism tutorial.

The result is a `PortableHaloCatalogue`, which behaves like any other catalogue but never refers back to the halo finder's files.

## The catalogue does not interpret its collaborators

Rather than the catalogue reaching into the mapper and the index lists — the very coupling this removes from tangos — each describes itself:

- `HaloNumberMapper.get_portable_state` / `from_portable_state`. The base implementation stores the halo numbers in index order, so an unknown or third-party mapper still transfers correctly; the built-in mappers override it to describe themselves exactly. `MonotonicHaloNumberMapper` names its class explicitly rather than via `type(self)`, so a subclass which does not override the method is rebuilt as the class whose description was actually stored.
- `HaloParticleIndices.get_portable_state` / `from_portable_state`, carrying the index list, the boundaries and `num_missing_particles` when present.

## What travels

The halo numbering; the particles belonging to each halo; the missing-particle counts, so halos which are incomplete under partial loading still raise `IncompleteHaloError` at the far end rather than quietly returning too few particles; whether the catalogue can determine completeness at all; and the halo finder properties from `get_properties_all_halos`, covering plain arrays, arrays with units, and ragged per-halo collections such as `children`. Anything that cannot be expressed portably is omitted with a warning.

Two things about completeness are worth review, since they are what connects this to the two PRs before it:

- **`_can_determine_completeness` is carried.** A catalogue whose membership is an array covering only the loaded particles (a `.grp` file, HOP output) cannot tell that a halo extends beyond them. Without the flag, the transferred copy would report every halo as complete, and without the warning explaining why — the transfer would appear to have created knowledge that does not exist.
- **`PortableHaloCatalogue` sets `_uses_file_position_addressing = False` explicitly.** Membership arrives as offsets into the snapshot the state came from, already resolved; nothing is expressed relative to a file, so the refusal that protects file-position catalogues from partially loaded snapshots must not fire here. The real requirement is that the simulation presents the same particles in the same order, which the caller asserts by passing it.

Properties available only one halo at a time — catalogues implementing `get_properties_one_halo` without `get_properties_all_halos` — are not included, and a catalogue which cannot expose the particles of all its halos raises `TypeError` rather than producing a state that silently omits them.

## Naming in `pynbody.array.shared`

Since this pulls the shared-array transfer API towards public use, its names are now descriptive (review suggestion, done in 6298448):

| was | now |
|---|---|
| `_deconstructed_shared_array` | `SharedArrayReference` |
| `pack` | `to_shared_reference` |
| `unpack` | `from_shared_reference` |

The old names remain as module-level aliases bound to the same objects, so existing tangos keeps working — including the underscore-prefixed class name, and including `isinstance` checks against it, since it *is* the new class. `test_shared_reference_legacy_names` asserts identity rather than equivalence, so the guarantee would fail loudly if a wrapper were ever introduced.

## Tests

`tests/halos_test.py` covers the round trip through `pickle`, to stand in for a real transfer, and asserts via `_assert_portable` that a state contains nothing but arrays, primitives and nested dicts of the same — so a state that quietly acquired an unportable object would fail. On top of that: every built-in mapper and an unknown mapper falling back to the base implementation; properties of each kind, including units and ragged `children`; `map_arrays` reaching every array; a real transfer through shared memory to a spawned subprocess; incompleteness surviving the transfer; the indeterminate-completeness case still warning afterwards; and reconstruction onto a view, which must not be refused. `tests/hbtplus_test.py` exercises the round trip on real HBT+ data.

## Notes

- **Format version.** `PORTABLE_STATE_VERSION` is 1, and stays 1: this module has never been released, so the completeness capability is part of the format's definition rather than an addition to a format anyone holds states in. `from_portable_state` refuses a state declaring a version it does not understand.
- This branch was brought up to date by merging master rather than rebasing, now that #1009 has landed: the four commits it was built on are ancestors of master, so it carries exactly one commit of its own and needed no history rewrite. The only conflict was two blocks of tests appended at the same point in `tests/halos_test.py`; both are kept.
- The transfer material has moved out of the halos tutorial, which was the wrong altitude for it, to the bottom of the parallelism tutorial; the halos tutorial links there. It became static `sourcecode` blocks in the move, since that document has no live IPython session.
- Drive-by: the test catalogue standing in for the indeterminate-completeness case was still described as the one that identifies particles by position, which after #1011 is the separate `_uses_file_position_addressing` category. Description corrected.

https://claude.ai/code/session_017aXdQBnpZpUD4Lo74m4Jtk


docs/tutorials/halos.rst
```diff
@@ -293,6 +293,34 @@ a function of mass in this halo catalogue:
  @savefig masses_vs_vels.png width=5in
  In [21]: plt.loglog()
 
+Transferring a catalogue to another process
```
apontzen: This does not belong here. The information is useful but is too detailed for an introductory tutorial. Place it instead at the bottom of the parallelism document

docs/tutorials/halos.rst
```diff
@@ -293,6 +293,34 @@ a function of mass in this halo catalogue:
...
+
+ In [25]: h_recreated[0]['vel'].std()
+
+The recreated catalogue offers the halo membership and finder-calculated properties of the original, but never
+touches the halo finder's files. Note that the snapshot it is attached to must have the same particle ordering
+as the one the state was generated from, since halo membership is stored as offsets into the snapshot.
+
 .. note::
 
     Pynbody includes infrastructure for analysing large simulations and halo catalogues using parallel processing.
```
apontzen: include a link to the parallelism document

pynbody/halo/portable.py
```diff
@@ -0,0 +1,245 @@
...
+        if self._original_class_name is None:
+            return super().__repr__()
+        else:
+            return f"<{type(self).__name__} from {self._original_class_name}, length {len(self)}>"
+
+
+def map_arrays(state, mapper, types=np.ndarray):
+    """Return a copy of a portable state, with every numpy array in it replaced by ``mapper(array)``.
+
+    This is a convenience for code that transfers a state elsewhere, such as a parallel task manager copying
```
apontzen: such a manager won't be *copying* the arrays into shared memory. They should already be in shared memory if pynbody has had shared memory enable. The manager's role will be to transmit the shared memory filename. This level of detail does not need to be in the docstring, just say "This is a convenience for code that transfers a state elsewhere e.g. via shared memory" or something like that.

pynbody/halo/portable.py
```diff
@@ -0,0 +1,245 @@
...
+    the arrays into shared memory. It saves the caller from having to know the structure of the state, let
+    alone the meaning of the individual arrays. For example, to transfer a catalogue using pynbody's own
```
apontzen: Getting far too detailed. Remove the example.

tests/halos_test.py
```diff
@@ -446,6 +453,369 @@ def test_unmatched_iords_in_partial_snapshot_are_incompleteness(snap_with_iords)
...
+    h_recreated = pynbody.halo.HaloCatalogue.from_portable_state(state, f)
+    assert (h_recreated[1].get_index_list(f) == h[1].get_index_list(f)).all()
+
+
+def _rebuild_catalogue_in_subprocess(packed_state, result_queue):
+    """Rebuild a catalogue that has been transferred through shared memory, and report on what it contains"""
+    import pynbody
+    from pynbody.array.shared import _deconstructed_shared_array, unpack
+
+    state = pynbody.halo.portable.map_arrays(packed_state, unpack, types=_deconstructed_shared_array)
```
apontzen: Why _deconstructed_shared_array? Since this is being pulled towards more public exposure, should it be instead named something like SharedArrayReference (or something even more descriptive if you can think of it)? I am giving permission to change pynbody.array.shared if this seems like a good idea, as well as this test and others that depend on it. 

While you're at it, perhaps `pack` and `unpack` should be `to_shared_reference` and `from_shared_reference` or something like that. 

In all cases, rename the primary instance but make synonyms preserving the old names (including the one with the leading underscore) to ensure older versions of tangos won't break.

apontzen: Done in 6298448, taking all three suggested names:

| was | now |
|---|---|
| `_deconstructed_shared_array` | `SharedArrayReference` |
| `pack` | `to_shared_reference` |
| `unpack` | `from_shared_reference` |

I couldn't better `SharedArrayReference` — it is a reference to an array in shared memory and holds none of the data, which is exactly what the name says. `to_`/`from_shared_reference` read correctly in both directions: the function does not move anything into shared memory, it converts an array that is already there into a reference to it.

The old names are module-level aliases bound to the same objects, so `from pynbody.array.shared import _deconstructed_shared_array, pack, unpack` keeps working, and an `isinstance` check against the old class name still matches, since it *is* the new class. `test_shared_reference_legacy_names` asserts identity rather than mere equivalence:

```python
assert pyn_array.shared.pack is pyn_array.shared.to_shared_reference
```

which is the strongest available statement that older tangos cannot break, and would fail if someone later reintroduced a wrapper.

Everything in the tree now uses the new names — the two tests, the shared-memory section of the parallelism tutorial, and the internal uses in `array/shared` itself. Both renamed functions carry a `.. versionchanged:: 2.7.0` noting the old name remains, and the class a `.. versionadded::` saying the same.

---

docs/tutorials/halos.rst
apontzen: Moved to the bottom of the parallelism document in 6298448, and removed from here.

One change it needed on the way: this document has no live IPython session, only static `.. sourcecode:: python` blocks, so the executed `.. ipython::` blocks would have had no `h` or `s` to draw on and would have failed the docs build. They are now static blocks matching the rest of that document.

---

docs/tutorials/halos.rst
apontzen: Added in 6298448. The existing note at the foot of this tutorial already mentioned the parallel processing infrastructure without linking to it, so the link went there rather than as a new line:

> Pynbody includes infrastructure for analysing large simulations and halo catalogues using parallel processing, including a way to hand a loaded catalogue to another process; see the :ref:`parallelism document <parallelism>`.

Using the `parallelism` label already defined at the top of that document, matching how `performance.rst` refers to it.

---

pynbody/halo/portable.py
apontzen: Corrected in 6298448 — and thank you, the old wording was not merely too detailed but wrong about the mechanism. The docstring is now two sentences:

> This is a convenience for code that transfers a state elsewhere, e.g. via shared memory. It saves the caller from having to know the structure of the state, let alone the meaning of the individual arrays.

The point about what is actually transferred is reflected in the naming as well, following your other comment: `to_shared_reference` produces a reference to an array already in shared memory rather than putting it there.

---

pynbody/halo/portable.py
apontzen: Removed in 6298448; the docstring now stops after the two sentences and goes straight to the parameters.

The equivalent worked example survives once, in the new section at the bottom of the parallelism document, where that level of detail is in keeping with the surrounding material.

---

apontzen:
**Backwards compatibility with tangos: verified.**

tangos master (`90e79ee`, v1.12.0 — a version with no knowledge of the rename) against pynbody built from this branch at 6298448:

```
513 passed, 1 skipped in 118s
```

numpy 2.3.5, SQLAlchemy 2.0.52, sqlite backend. The skip is pre-existing and unrelated (`test_parallel_tasks_async.py:33: Async processing is currently switched off`).

A green suite would not on its own show that the renamed code was reached, so I instrumented `pynbody.array.shared` to record every call. During that passing run, across 44 processes:

| symbol | calls |
|---|---|
| `SharedArrayReference` (the renamed class) | 150 |
| `_shared_array_deconstruct` | 75 |
| `_shared_array_reconstruct` | 75 |

So the renamed class is genuinely constructed and passed through tangos' own transfer code, rather than the suite merely coexisting with the change.

One thing worth recording: tangos does not use `pack`/`unpack` at all — `transfer_array.py:39,44` calls the private `_shared_array_deconstruct`/`_shared_array_reconstruct`, which are not renamed. The alias that actually protects tangos is therefore `_deconstructed_shared_array`, the underscore-prefixed one; the `pack`/`unpack` aliases are belt-and-braces for other consumers.

Scope of the claim: this tests tangos 1.12.0, the direction that matters here. Releases predating 1.12.0 are untested, and are protected only insofar as they use the same entry points.

The subsequent commit on this branch is docstring-only (2.7.0 version markers), so it does not affect the above.

---

apontzen:
**The failing `docs/readthedocs.org:pynbody` check is not this PR's.** Recording why, since the head is otherwise ready.

[Build 34348268](https://app.readthedocs.org/projects/pynbody/builds/34348268/) ran 13 commands and died on the last one, before Sphinx started:

```
cd tests && python -c "import pynbody.test_utils as tu; tu.precache_test_data(extended=True)"
...
File ".../osfclient/models/file.py", line 64, in write_to
    raise RuntimeError("Response has status code {}.".format(response.status_code))
RuntimeError: Response has status code 403.
```

That is OSF refusing the test-data download. The GitHub Actions `fetch-testdata` jobs passed on the same commit because they restore from cache rather than fetching, so the 403 shows up only where the download is cold. Nothing in this PR touches test data or the docs build, and the previous commit (`8b636af3`) built fine 1½ hours earlier; the only change since is four docstrings.

I have no means to re-trigger a Read the Docs build from here (no token, and kicking it with an empty commit is not something I will do), so this needs either a Rebuild from the RTD UI or the next push to the branch.

Since the point of this commit was to fix version directives rendering *inside* numpydoc `Returns`/`Parameters` blocks, and RTD is the usual way to confirm that, I verified it independently with a local Sphinx build of the six affected docstrings:

| object | admonition | `Parameters` | `Returns` |
|---|---|---|---|
| `from_portable_state` | line 15 | 16 | 28 |
| `complete_keys` | 42 | 43 | — |
| `map_arrays` | 18 | 19 | — |
| `load` | 17 | 25 | 46 |
| `to_shared_reference` | 11 | 15 | 28 |
| `from_shared_reference` | 9 | 13 | 20 |

In every case the admonition now renders ahead of any field list, i.e. as its own "Added in version 2.7.0." / "Changed in version 2.7.0:" rather than as part of the return value or a parameter description.

---

PR #1011, opened 2026-08-12
Handle partially loaded snapshots correctly across all halo finders

apontzen:
Builds on the `complete_keys()` API from #1009 (now on master) and makes every halo finder behave sensibly when the snapshot has been partially loaded. There are two halves, because the finders fall into two categories.

## Finders which identify particles by ID

These can detect that a halo extends beyond the loaded particles, so they now do. Rockstar, Subfind (when reordering), AHF, VELOCIraptor and AdaptaHOP map their particle IDs with `_map_iords_to_fpos` / `_map_iords_to_fpos_one_halo`, so a missing particle is counted rather than raising `IllegalIordError`, and the affected halos are reported through the existing `complete_keys()` / `is_complete()` / `get_complete_mask()` API.

All five assembled their index lists by writing into a preallocated array using file-derived boundaries, which no longer holds once particles are dropped. A shared helper does the compaction:

```python
HaloCatalogue._assemble_particle_indices(mapped_iords_per_halo, num_halos, num_particles, sort=False)
```

It takes an iterable of `(file_positions, num_missing)` pairs, one per halo, compacts the index list, rebuilds the boundaries and records the counts. HBT+ is moved onto it, replacing its hand-rolled equivalent.

Two needed more than a call swap: VELOCIraptor's `__get_particle_indices_from_halo_index` mapped IDs internally, so it is now `__get_particle_ids_from_halo_index` returning raw IDs, with the mapping and the combined bound + unbound missing count done by the caller; and AHF maps inside `_ahf_to_pynbody_particle_ids`, so that and `_load_ahf_particle_block` now return `(ids, num_missing)`.

A particle ID which is absent because it was never loaded is indistinguishable from one absent because we have misread the catalogue's ID convention — unless the snapshot is complete, in which case nothing *can* be missing. `_map_iords_to_fpos` uses that: on a complete snapshot it raises `IllegalIordError` naming the likely cause, rather than reporting most of the halos as incomplete.

## Finders which identify particles by file position

A halo's membership here is expressed as offsets into the snapshot file, and if some particles were never loaded those offsets denote different particles. Making these work properly would mean exposing pynbody's file layout logic systematically to the halo module — per-`PartType` versus per-family addressing, concatenation across files, ramses' lack of a `LoadControl` — so instead these catalogues refuse to load, raising the new `PartialLoadingNotSupportedError`.

**This trade — silently wrong data becomes a hard error — is a deliberate decision, and is approved.** It is a behaviour change for anyone combining these finders with partial loading, so it wants a release note, but nothing that previously *worked* is lost. What it replaces:

| case | result before this PR |
|---|---|
| AHF halo 12, entirely inside the loaded region | 5294 particles returned, **0** of them correct |
| AHF halo 28 (contains gas) | `IndexError: index 872541 is out of bounds for axis 0 with size 853578` |
| SubFind halo 1, entirely inside the loaded region | 19153 particles returned, **500** not members of the halo at all |

Controlled by `HaloCatalogue._uses_file_position_addressing`, checked via `_refuse_if_snapshot_partially_loaded`. It applies to the four SubFindHDF classes, AHF unless it is using iords *and* the newer file format, Subfind where the snapshot is taken to be ordered, and — checked lazily, since it is only discoverable at mapping time — any catalogue falling back to sequential IDs because no `iord` array exists.

Two details worth reviewing:

- The check is in `__init__`, never in `_can_load`. `SimSnap.halos()` tests `_can_load` in priority order and then constructs without a try/except, so a refusal in `_can_load` would silently fall through to a *different* halo finder — the worst outcome, since the user would get a plausible-looking catalogue from the wrong code.
- AHF and SubFindHDF also check early, before they go looking for their files. Otherwise `f[:1000].halos()` reported "No candidate AHF catalogue found", which points at the wrong problem. For AHF the early answer is deliberately weaker than the final one, because with `use_iord=True` it still depends on the file format revision, which cannot be read until the files have been located — and for a view they cannot be located at all. `_determine_file_position_addressing` holds that logic in one place and is called at both points; `HaloCatalogue.__init__` re-runs the refusal with the final answer.

Catalogues built from an array of halo numbers per particle (`HaloNumberCatalogue`, and hence HOP) are in neither category: the array covers exactly the particles which were loaded, so it stays self-consistent. It simply cannot tell that a halo extends beyond them, so it keeps `_can_determine_completeness = False` and warns if asked about completeness. After this PR that is the only case where the warning fires — for the file-position finders, if the catalogue exists at all then the snapshot is complete, and so is every halo.

## Identifying a partially loaded snapshot

The refusal is only as trustworthy as the test behind it, so `SimSnap` gains a public, tested predicate:

```python
f.is_partially_loaded()
```

True whenever the snapshot holds less than its file, however that came about: particle selection within the files opened (`take=`, `take_region=`, swift's `take_swift_cells`), selection of whole files (ramses `cpus=`, `with_gas=False` or `maxlevel=`, or being pointed at one file of a multi-file HDF5 set), or a view such as `f[:100]` or `f.dm`. Views count because they may also reorder particles. `CopyOnAccessSimSnap` inherits the state of whatever it wraps, which matters because that is the case tangos exercises.

Filling the gaps needed a few small loader changes: `nchilada`, `grafic` and `ascii` accepted `take` without recording it, ramses partial-loads via `cpus=` which is not a `take` at all, and being handed one file of a multi-file set was previously indistinguishable from a single-file snapshot (`_GadgetHdfMultiFileManager` now compares the file's declared `NumFilesPerSnapshot` against what it opened).

Underneath the predicate are two attributes, kept separate because they differ in what they permit rather than in how the snapshot came to be incomplete. The question both answer is whether an index into the snapshot still addresses the same particle in `filename`:

- `partial_load` — it does not. `load_copy()` re-opens `ancestor.filename` and passes `take=get_index_list(ancestor)`, and the tipsy write-back paths re-read by index too, so both must refuse. Set by the `take` loaders, and by ramses for a cpu subset or dropped gas, since its `filename` is the whole output directory.
- `incomplete_file_set` — it does. `filename` is the single file that was opened and the snapshot holds exactly its contents in its order, so re-reading by index remains valid; the snapshot simply does not contain all the particles other tools expect the output to have.

Merging them broke `test_partial_load_mass_in_header`, which loads file 0 of a gadget set and asserts `f[::2].load_copy()['mass']` matches `f[::2]['mass']` — a legitimate operation that a merged flag would refuse.

Finally, `SimSnap.halos()` now adds a hint when no catalogue could be loaded and the snapshot is partial, since a subsnap's mangled filename also defeats catalogue auto-detection — a pre-existing wrinkle that this change makes more visible.

## Tests

- `tests/subfind_test.py` — real partial-loading coverage for an ID-based finder: a catalogue built on `f[len(f)//2:]` (the second half, so the iords are not sequential and mapping actually happens) reports 1176 of 2853 groups complete, incomplete groups raise `IncompleteHaloError`, retained particles plus missing count equal the finder's `group_len`, and complete groups match the full-snapshot particle lists exactly.
- `tests/velociraptor_test.py` — the same over partially loaded SWIFT, including the unbound particle list, with `test_velociraptor_id_dtype_differs_from_snapshot` as an explicit control so the tests cannot quietly stop exercising the int64-catalogue-against-uint64-`iord` path.
- `tests/ahf_halos_test.py`, `tests/subfindhdf_gadget4_test.py` — refusal tests for `take=`, for views, for a copy-on-access wrapper, and through `halos()`, each paired with a positive control that the same catalogue loads fine from a complete snapshot. `test_ahf_loads_from_complete_snapshot` asserts `use_iord` is False, so the tests cannot quietly stop exercising the offset path; `test_ahf_refuses_partially_loaded_snapshot_using_old_format_iords` covers the one path where only the recheck in `HaloCatalogue.__init__` can refuse.
- `tests/snapshot_test.py` — tests pinning `is_partially_loaded` across every origin above.
- `tests/halos_test.py` — an empty selection (`f[:0]`) yields incomplete halos rather than a reduction error, plus the indeterminate-completeness warning.
- `tests/utils_test.py` — `binary_search` across every combination of int32/int64/uint64 for needle, haystack and sorter, and uint64 values above the int64 range.

## Notes

- **Version target.** 2.6.0 shipped while this work was in progress, so `is_partially_loaded` is marked `.. versionadded:: 2.7.0`, matching the rest of the API.
- Review findings so far have all been addressed — see the resolved threads for which commit fixed each. The `.gitignore` additions that were in an earlier revision of this branch have been dropped, as unrelated to it.

## One residual ambiguity

On a snapshot which really is partially loaded, a misread ID convention still cannot be distinguished from genuinely absent particles, so it presents as halos being incomplete rather than as an error. The loud error is available only when the snapshot is complete, which is where `test_ahf_wrong_id_convention_is_reported` pins it. This is inherent to ID-based detection rather than a gap in the implementation.


apontzen:
pre-commit.ci autofix

pynbody/snapshot/simsnap.py
```diff
@@ -605,6 +620,27 @@ def ancestor(self) -> SimSnap:
...
+        :class:`~pynbody.halo.HaloCatalogue`.
+
+        Note that a view which happens to contain all the particles of its ancestor is still reported as
+        partially loaded, since it may present them in a different order.
+
+        .. versionadded:: 2.6.0
+
+        """
+        ancestor = self.ancestor
+        return self is not ancestor or ancestor.partial_load or ancestor.incomplete_file_set
```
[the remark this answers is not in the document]
apontzen: Fixed in 7497a8c. `CopyOnAccessSimSnap.__init__` now inherits the partial-loading state of whatever it wraps:

```python
self.partial_load = base is not base.ancestor or base.ancestor.partial_load
self.incomplete_file_set = base.ancestor.incomplete_file_set
```

A view counts as a selection within the files opened, hence `partial_load`; whether the whole file set was opened is a property of the files themselves, hence the separate attribute. This matters more than the general case, since wrapping a partially loaded snapshot is precisely what tangos does.

---

pynbody/snapshot/ramses.py
```diff
@@ -447,6 +447,10 @@ def __init__(self, dirname, cpus=None, maxlevel=None, with_gas=True, force_gas=F
             self._cpus = cpus
         else:
             self._cpus = list(range(1, self.ncpu + 1))
+
+        # ramses partial-loads by selecting whole cpu files; anything less than all of them means the
+        # snapshot does not contain all the particles in the output
+        self.partial_load = sorted(self._cpus) != list(range(1, self.ncpu + 1))
```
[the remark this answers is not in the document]
apontzen: Fixed in 7497a8c. Both options are now recorded as partial loads, guarded on there being gas on disk in the first place so that `with_gas=False` on a gas-free output is not flagged:

```python
if gas_on_disk and (not has_gas or maxlevel is not None):
    self.partial_load = True
```

You were right that it mattered: in the test output either option leaves out all 1.4M gas particles while claiming to be complete, so AHF constructed happily and then failed with an `IndexError` from inside the subsnap machinery. `maxlevel` is taken at face value rather than checking whether it actually truncated anything, since establishing that would mean reading the AMR depth that `maxlevel` exists to avoid reading.

---

pynbody/halo/__init__.py
```diff
@@ -501,6 +557,50 @@ def _map_iords_to_fpos_one_halo(self, iords, halo_number=None) -> NDArray[int]:
             raise IncompleteHaloError(halo_number, num_missing)
         return fpos
 
+    def _assemble_particle_indices(self, mapped_iords_per_halo, num_halos, num_particles,
+                                   sort=False) -> HaloParticleIndices:
+        """Assemble the index list for all halos, given their particles' file positions.
```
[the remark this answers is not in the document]
apontzen: Fixed in 7497a8c. `make_iord_to_offset_mapper` now short-circuits before the reductions:

```python
if len(iord) == 0:
    return IordToOffsetDense(iord, max_iord=-1)
```

An empty dense lookup table reports every query as missing, which is exactly what an empty snapshot should do, so the halos come back incomplete rather than the reduction raising. Two regression tests cover it: `test_empty_iord_to_pos_map` for the mapper itself, and `test_catalogue_against_empty_selection` for an ID-based catalogue built on `f[:0]`.

---

pynbody/halo/__init__.py
```diff
@@ -187,17 +191,49 @@ class HaloCatalogue(snapshot.util.ContainerWithPhysicalUnitsOption,
...
-      ``num_missing_particles``, so that loading all halos does not fail, while an
+      particles. The counts are passed to :class:`details.particle_indices.HaloParticleIndices` as
+      ``num_missing_particles``, most easily by assembling the index list with
+      :meth:`_assemble_particle_indices`, so that loading all halos does not fail, while an
       :class:`IncompleteHaloError` is still raised if an affected halo is accessed. Those counts are also what
       :meth:`complete_keys` reports to the user, so no further work is needed to support it.
+    * If, on the other hand, the format identifies its particles by position within the snapshot rather than
+      by ID, missing particles cannot be detected at all. Such subclasses should set the class attribute
+      :attr:`_can_determine_completeness` to False, so that users are warned instead of being told that every
+      halo is complete.
```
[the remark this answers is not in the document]
apontzen: Fixed in 7497a8c — this was the most useful of the four, since the guidance was actively misleading subclass authors. The class docstring now describes the three categories separately: ID-based finders pass `num_missing_particles`; file-position finders set `_uses_file_position_addressing = True` so construction is refused; and only membership arrays defined over the loaded particles (a halo number per particle) set `_can_determine_completeness = False`, which warns. Each attribute's own docstring says which category it is for, and `complete_keys` names the right one.

---

docs/tutorials/parallelism.rst
```diff
@@ -127,6 +127,17 @@ particles to a halo which have not been read from disk. Such a halo is *incomple
 raises an :class:`~pynbody.halo.details.particle_indices.IncompleteHaloError` rather than silently returning
 too few particles.
 
+This is possible only for halo finders which identify their particles by ID. Some formats instead express
+halo membership as positions within the snapshot file, which are meaningless unless every particle is
+present. Those catalogues refuse to load against a partially loaded snapshot, raising a
+:class:`~pynbody.halo.details.particle_indices.PartialLoadingNotSupportedError`, rather than silently
```
apontzen: "rather than silently returning the wrong particles" is not helpful - end the sentence so that it states what it does, not what it doesn't do. Also unhelpful is the sentence starting "At the time of writing this affects..." - the "at the time of writing" deflects the guarantee so that the statement becomes worse than useless. Let's just leave it out.

pynbody/halo/ahf.py
```diff
@@ -120,7 +127,11 @@ def __init__(self, sim, filename=None, make_grp=None, get_all_parts=None, use_io
 
         number_mapper = self._setup_halo_numbering(halo_numbers)
 
-        super().__init__(sim, number_mapper)
+        # only the new format maps its particle IDs; otherwise they are interpreted as offsets within the
+        # snapshot's families, which is only meaningful if all the particles are present
+        self._uses_file_position_addressing = not (self._use_iord and self._is_new_format)
```
apontzen: this is contradictory / more restrictive than line 105 above; the intended refusal won't be issued  if self._use_iord is True but self._new_format is False. Seems problematic; shouldn't this be inferred just once? Or at the very least, the refusal barrier should be run again with this tighter restriction once it's known

pynbody/snapshot/ramses.py
```diff
@@ -447,16 +447,26 @@ def __init__(self, dirname, cpus=None, maxlevel=None, with_gas=True, force_gas=F
...
```
apontzen, line 453: should this be setting `incomplete_file_set` rather than `partial_load`? May be academic as I don't think ramses supports `take`, but it might be more accurate if later ramses starts supporting `take` (which is a strong possibility)

docs/tutorials/parallelism.rst
apontzen: Both done in 15a94520. The paragraph now ends at what happens:

> Those catalogues refuse to load against a partially loaded snapshot, raising a `PartialLoadingNotSupportedError`.

and the list of affected finders is gone.

One consistency point I have left alone: the paragraph immediately above has the same shape — "raises an `IncompleteHaloError` rather than silently returning too few particles". That text is already on master, from #1009, so I have not touched it here; say the word and I will apply the same treatment.

---

pynbody/halo/ahf.py
apontzen: The refusal *is* issued in that case, but only because `HaloCatalogue.__init__` re-runs the check — which is far too subtle to rely on the reader spotting, so I have restructured it in 15a94520.

To take the mechanism first: `HaloCatalogue.__init__` opens with `self._refuse_if_snapshot_partially_loaded(sim)` (`pynbody/halo/__init__.py:246-247`), so the assignment on the line you commented on is followed immediately by a check using it. With `use_iord=True` and the older format the value becomes `True` there, and `super().__init__` raises. Nothing between the two lines touches the snapshot, so there is no earlier failure to trip over.

So the two assignments are not contradictory — the earlier one is deliberately *weaker*, refusing only what is certain regardless of how the format revision turns out, and the later one is authoritative. That relationship was implicit, and your reading is the natural one.

On inferring it once: the authoritative answer genuinely cannot be computed at the earlier point, because it needs the format revision, which needs the catalogue files to have been located — and for a view such as `f[:100]` they cannot be located at all, since the filename no longer corresponds to the snapshot. That is exactly the case the early check exists for. So I have kept two calls but moved the logic into one place:

```python
def _determine_file_position_addressing(self, format_revision_known: bool) -> None:
    if format_revision_known:
        self._uses_file_position_addressing = not (self._use_iord and self._is_new_format)
    else:
        self._uses_file_position_addressing = not self._use_iord
```

with a docstring stating why it is called twice and that the second answer is never less restrictive than the first, and the call sites reading `format_revision_known=False` / `True`.

The more useful outcome of your question: **the combination you identified had no test.** `use_iord=True` against the older format is the one path where the refusal can only come from the recheck, and nothing exercised it. `test_ahf_refuses_partially_loaded_snapshot_using_old_format_iords` now does; since the test data is new-format, it subclasses to force `_is_new_format = False`. If that test ever fails, the recheck has been lost.

---

pynbody/snapshot/ramses.py
apontzen: `partial_load` is the right one here, and it stays right even if ramses gains `take` — but the docstrings did not make that clear, so I have reworded them in 15a94520.

The operative difference between the two attributes is not really "within a file" versus "whole files"; it is whether an index into the snapshot still addresses the same particle in `filename`. That is what the consumers actually depend on: `load_copy` (`simsnap.py:915`) re-opens `ancestor.filename` and passes `take=self.get_index_list(self.ancestor)`, and the tipsy write-back paths refuse for the same reason.

For gadget-HDF pointed at one file of a set, `filename` *is* that file and the snapshot holds exactly its contents in its order, so re-reading by index is valid — hence `incomplete_file_set`, and hence `test_partial_load_mass_in_header` passing, which is what stopped me merging the two attributes in the first place.

For ramses, `self._filename` is the output *directory* (`ramses.py:427`), so re-loading it yields every cpu file. Indices derived from a `cpus=[1]` snapshot would then address the wrong particles. Marking it `incomplete_file_set` would license precisely the operation that must be refused, so `partial_load` it is.

The docstrings now define both by that invariant rather than by the mechanism which set them, which also answers your forward-looking point: if ramses starts supporting `take`, `partial_load` will cover both routes and still mean exactly what it says.

---

PR #1014, opened 2026-08-14
Add user-defined pairwise SPH operations to KDTree

apontzen:
## Motivation

Expressing a pairwise SPH operation — artificial viscosity, artificial conduction, or anything else summing over neighbour pairs — currently means looping over `KDTree.nn()` from python. Measured on 200k particles at 64 neighbours, that costs ~400 ns per neighbour against ~4 ns for the equivalent C++ pass, and it leaves the caller to rebuild the pair list, the periodic wrapping and the kernel normalisation by hand.

These operations can't be assembled from a fixed set of linear SPH sums: the pair function is genuinely nonlinear in the pair, containing terms like `sqrt(2|P_i-P_j|/(rho_i+rho_j))` and `min(mu_ij, 0)`. So the primitive has to accept user code, and the only real design question is granularity — per-pair would mean ~1e8 python calls, per-particle leaves numpy overhead dominating on length-32 arrays. This works on **blocks of pairs**, defaulting to 2^18 at a time.

## What's added

**`KDTree.pair_blocks(mode, blocksize)`** yields `(i, j, dx, r)` as flat read-only numpy arrays, with `dx` already periodic-wrapped. Note `i` and `j` are index *arrays*, one entry per pair, so per-particle quantities are reached by fancy indexing.

- `'gather'`: ordered pairs with `r <= 2 h_i` — the neighbour set behind the density estimate.
- `'symmetric'`: each unordered pair exactly once, `r <= max(2 h_i, 2 h_j)`, canonicalised to `i < j` — what operators involving *both* smoothing lengths require.

**`KDTree.pair_reduce(func, ...)`** accumulates a callback over those pairs. The callback returns `contrib_i` or `(contrib_i, contrib_j)`; pynbody does the `bincount` scatter-add.

**Vectorised `KernelBase.value(r, h)` / `.gradient(r, h)`, and `KDTree.kernel`**, so a callback can evaluate the same kernel the C++ is using rather than re-deriving its normalisation.

Artificial conduction then reads:

```python
m, rho, p, u, h = (f.g[k] for k in ('mass','rho','p','u','smooth'))
W = f.g.kdtree.kernel

def conduction(i, j, dx, r):
    mu  = np.einsum('kl,kl->k', vel[j] - vel[i], dx) / r
    v_d = 0.5 * (np.abs(mu) + np.sqrt(2*np.abs(p[i]-p[j]) / (rho[i]+rho[j])))
    g   = -(W.gradient(r, h[i])/rho[i] + W.gradient(r, h[j])/rho[j])
    c   = v_d * (u[j] - u[i]) * g
    return m[j] * c, -m[i] * c

du_dt = f.g.kdtree.pair_reduce(conduction)
```

Because each pair is visited once and both ends are accumulated from that single visit, an antisymmetric callback conserves energy **as an algebraic identity** rather than approximately — measured residual ~6e-19.

## Implementation notes

**No `hMax` needed on `KDNode`.** Symmetric mode falls out of a purely local test: while walking particle `a` and finding neighbour `b`, emit iff `a < b` or `r > 2 h_b`. In the second case `b`'s own walk cannot reach `a`, so nobody else emits it; in the first, the lower index breaks the tie. No global de-duplication, no communication between walks, and the node struct (mirrored in the python `KDNode` dtype and in `serialize`/`deserialize`) is untouched.

**Membership is tested on `r`, not `r^2`.** Since `h` is defined as half the distance to the nth neighbour, `r == 2h` is hit *exactly* for one neighbour of every particle — measured at exactly 400 boundary pairs on 400 particles. Deciding on `r^2` strands those pairs on the wrong side by an ulp. The boundary is inclusive, matching `smBallGather`, so a gather over `nSmooth` neighbours really does return `nSmooth` of them.

## Drive-by bug fix

`WendlandC2Kernel::gradient` in `kernels.hpp` omitted the `21/16` prefactor that its `operator()` includes. `smDivQty`/`smCurlQty` apply the same `1/(pi h^n)` factor to both, so **`sph_divergence` and `sph_curl` were wrong by 16/21 ≈ 0.762 whenever the Wendland kernel was selected**; the cubic spline was unaffected, which is why no existing test caught it. Measured ratio against an independently-verified python gradient was exactly 1.312500 before the fix, 1.000000 after.

This changes numerical output for anyone using `sph_divergence`/`sph_curl` with Wendland.

## Testing

`tests/user_sph_operations_test.py`, 86 tests. Every assertion runs against **both** the C++ path and a brute-force O(N²) reference implementation defined in the file, which serves as the specification. Coverage includes the callback contract, blocksize invariance and determinism, both pair sets, periodic minimum-image displacement, accumulation semantics (including the `bincount` vs `out[i] +=` trap), the conservation and positive-definiteness invariants, and cross-checks reproducing pynbody's own `rho`, `sph_divergence` and `sph_curl`.

Where the two implementations disagree on the `r == 2h` boundary — irreducibly, since two independent float evaluations of the same distance differ in the last bit — the tests assert the disputed pairs carry exactly zero kernel weight, so nothing physical can depend on them.

The new regression test was verified to fail against a deliberately re-broken build.

Version bumped to 2.6.0, with `.. versionadded:: 2.6.0` on each new public entry point.

https://claude.ai/code/session_01YGC9utrm2V3Tbu6o6GwgK5

### DIFF
--- a/pynbody/__init__.py
+++ b/pynbody/__init__.py
@@ -69,6 +69,6 @@ from .snapshot import load, new
 
 derived_array = snapshot.simsnap.SimSnap.derived_array
 
-__version__ = '2.5.0'
+__version__ = '2.6.0'
 
 __all__ = ['load', 'new', 'derived_array']
--- a/pynbody/kdtree/__init__.py
+++ b/pynbody/kdtree/__init__.py
@@ -29,6 +29,21 @@ KDNode = np.dtype([
     ('pUpper', np.intp)
 ])
 
+def _scatter_add(out, index, contrib):
+    """Perform ``out[index] += contrib``, summing over repeated indices.
+
+    ``np.add.at`` would do this, but is very slow; ``out[index] += contrib``
+    is wrong, since it keeps only one contribution per repeated index.
+    """
+    npart = out.shape[0]
+    if out.ndim == 1:
+        out += np.bincount(index, weights=contrib, minlength=npart)
+    else:
+        for k in range(out.shape[1]):
+            out[:, k] += np.bincount(index, weights=contrib[..., k],
+                                     minlength=npart)
+
+
 class KDTree:
     """KDTree can be used for smoothing, interpolating and geometrical queries.
 
@@ -75,6 +90,12 @@ class KDTree:
     PROPID_QTYDISP_ND = 6
     PROPID_QTYDIV  = 7
     PROPID_QTYCURL = 8
+
+    PAIR_MODE_GATHER = 0
+    PAIR_MODE_SYMMETRIC = 1
+
+    _pair_modes = {'gather': PAIR_MODE_GATHER,
+                   'symmetric': PAIR_MODE_SYMMETRIC}
 
     def __init__(self, pos, mass, leafsize=32, boxsize=None, num_threads=None, shared_mem=False):
         """Create a KDTree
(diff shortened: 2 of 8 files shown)


apontzen:
Thanks — three of these were real, all fixed in 8d30626.

**Docstrings** (`pair_blocks`, and the `PairContext` comment in `smooth.h`): correct, and my error. Membership moved from `r²` to `r` with an inclusive boundary during development; the code and tests were updated but these two prose descriptions weren't. Also right that the symmetric emit rule is `r > 2h_b` rather than `>=`, so equality falls to `b`'s walk.

**Integer `dtype`**: confirmed — `pair_reduce(..., dtype=int)` raised `UFuncTypeError`. Fixed, but not by casting the `bincount` result per block: that truncates each partial sum independently, so the answer starts depending on `blocksize`, which `pair_blocks` documents as an implementation detail.

```
per-block cast, blocksize=1000 -> 3
per-block cast, blocksize=3    -> 2      (exact: 6*0.6 = 3.6 -> 3)
```

Accumulation now stays in float64 and converts once on return. The new test asserts total truncation is bounded by one unit rather than growing with the block count, and was verified to fail against the per-block variant. It deliberately doesn't claim exact agreement between blocksizes — float64 partial sums differ in the last bit with summation order, which can tip a value either side of an integer.

**`smBallGatherStoreResultInList`**: right that it returns `particleIndex + 1` where `smBallGather` expects the next write position. Pre-existing and inert (the only caller discards the count and reads `smx->result`, which is why `test_particles_in_sphere[100000]` passes), but fixed since it's a trap for any future caller.

PR #1015, opened 2026-08-15
Fix SPH smoothing on large and periodic snapshots, and add a volume/mass weighting option

apontzen:
Correctness and performance fixes to the KDTree SPH operations, plus a new `weighting` option for them.

Most of these surfaced while exercising the pairwise API added in #1014 on large cosmological snapshots. Several are pre-existing and affect `sph_mean`, `sph_dispersion`, `sph_divergence`, `sph_curl` and `rho`, not only the new code.

**Behaviour change to flag:** `sph_divergence` and `sph_curl` return different values for particles near a periodic boundary, because they were previously wrong there. Everything else leaves existing results untouched.

---

### The neighbour buffer could overflow when smoothing lengths come from a snapshot

The buffer used to collect a particle's neighbours is sized as `config['sph']['smooth-particles'] + RESMOOTH_SAFE`, 532 by default. That is meaningful only when the smoothing lengths came from pynbody's own nearest-neighbour search, which by construction puts a known number of particles inside each kernel. When `h` is instead read from the snapshot, as for any SWIFT or GADGET file carrying its own smoothing lengths, a kernel can enclose arbitrarily many particles and the gather overflows:

```
RuntimeError: Buffer overflow in smoothing operation
RuntimeError: Buffer overflow while gathering neighbour pairs
```

This affects `populate` and therefore `sph_mean`, `sph_dispersion`, `sph_divergence`, `sph_curl` and `rho`, as well as the new `pair_blocks`. On a cosmological hydrodynamics snapshot it triggers within half a second.

The local density does not bound the problem. Estimating the worst case as `(4/3)π(2h)³ρ/m` gives a maximum of 142 neighbours across 42M particles, where the true figure exceeds 532: `ρ_i` is a kernel-weighted average dominated by nearby particles, while the count scales with volume, so a particle whose `h` was set in a sparse region and which now sits beside a dense one exceeds it. The measured mean was 78 neighbours within `2h`, with a sampled maximum of 327.

The buffer now doubles and the gather is retried, which is amortised O(1) and leaves it large enough for subsequent particles. Each thread owns its own buffer, so this needs no synchronisation.

### `sph_divergence` and `sph_curl` ignored periodicity

`smDivQty` and `smCurlQty` obtained the pair displacement by differencing the stored positions, while the ball-gather that found the neighbour works in the periodic minimum image. For a pair straddling a box face the two disagree by a box length, so the displacement used bore no relation to the separation the neighbour was selected on.

Every particle within `2h` of a box face was affected. On a uniform periodic test that is 3158 of 8000 particles, with 100% of the disagreements adjacent to a face; on a non-periodic distribution there is no discrepancy at all. The existing `test_div_curl_smoothing` uses a zoom simulation whose particles lie far from the box faces, which is why it neither caught this nor changes now.

The gather already records the wrapped displacement, so both routines now read it from there.

### `get_arrayref` segfaulted when the array was unset

`Py_INCREF` was called before the `NULL` check, so requesting an array that had never been set dereferenced `NULL` instead of returning `None`.

## Performance

`pair_reduce`'s accumulation used `np.bincount(..., minlength=N)`, which allocates and fills a temporary as long as the output on every call. For a large snapshot that dominates everything else — at 42M particles it moves 340 MB per block per end:

```
bincount:    122 ms and 100 ms per block
np.add.at:   0.8 ms and 0.7 ms per block
```

`np.add.at` is the slower of the two only while the output is comparable in length to the block, which is the regime the previous code assumed. The choice is now made on that ratio, with `bincount` retained where it wins. This is worth 2.7x on a large snapshot and nothing is given up on a small one.

## New: `weighting='volume'` or `'mass'`

`sph_mean`, `sph_dispersion`, `sph_divergence` and `sph_curl` divide each neighbour's contribution by that neighbour's own density, `m_j/ρ_j`. Dividing by `m_j/ρ_i` instead is equally standard and is the form SPH codes use internally. Both are now available; the default is unchanged.

The two are not competing estimates of one quantity. They converge to different quantities:

| | limit | consistent with |
|---|---|---|
| `'volume'` (default, `m_j/ρ_j`) | `∫ q W dV` | the continuum smoothing integral, of which it is the direct quadrature |
| `'mass'` (`m_j/ρ_i`) | `∫ q ρ W dV / ∫ ρ W dV` | the discrete mass distribution: the weights `m_j W_ij` sum to `ρ_i` by definition |

Interpolating a field correlated with density shows the difference directly: the ratio between them is 1.088 for `q = ρ`, 0.938 for `q = 1/ρ`, and 1.015 for an uncorrelated field.

Naming them for what they average over makes the properties follow from the name. A mass-weighted average is a true weighted mean, so it reproduces a constant field exactly and can never return a value outside the range of its input:

```
interpolating a constant 3.25:
  'mass'     3.250000   max error 3e-15
  'volume'   3.229914   max error 1.2
```

A volume-weighted average makes no assumption about density being constant across a kernel, and is the faithful discretisation of the smoothing integral.

For a divergence the mass form is the one that satisfies the continuity equation exactly (verified to `6.4e-16`), and hence the one appearing in the SPH equations of motion. It is therefore what is needed to reproduce a velocity divergence as the simulation itself computed it: against the `VelocityDivergences` stored in a SWIFT snapshot, correlation is 0.94 with `'mass'` and 0.41 with `'volume'`.

`rho` and `smooth` take no such weighting and are unaffected — `smDensity` is what defines ρ.

## Documentation

Several docstrings were inaccurate rather than merely unclear:

- `pair_blocks` stated that accessing `f['smooth']` fixes the pair set. That holds only where pynbody derives the smoothing lengths; where they come from the file, reading them never involves the tree and they must be associated with it explicitly. The error message now distinguishes the two cases.
- The `nsmooth` argument to `sph_mean`, `sph_dispersion`, `sph_divergence` and `sph_curl` was described as the number of neighbours to use. It does not select the neighbours, which follow from the smoothing lengths — `sph_mean(q, 16)` and `sph_mean(q, 256)` return bit-identical answers. It sizes the buffer and sets the neighbour count that kernels such as Wendland C2 are normalised for.
- `populate` documented four of the six operations it accepts.
- The `KDTree` class docstring gave `f.build_kdtree()` as a worked example; the method is `build_tree()`.
- `sph_divergence`'s parameter and return descriptions were copy-pasted from `sph_curl` and named the wrong operation.
- `sph_mean` and `sph_dispersion` wrote `q_i` inside a sum over `j`, and the dispersion gave `(q_i - r_i)²` where the code forms the difference against each neighbour's value.

Remaining changes are consistency: indices are defined before use, `pair_reduce` no longer overloads `i` and `j` for both the ends of a pair and the particle being accumulated into, and it now states that the callback runs on a single thread.

## Testing

217 tests pass. Each new test was checked against the unfixed code to confirm it fails there:

- `test_pair_blocks_handles_far_more_neighbours_than_the_default_buffer`
- `test_smoothing_operations_handle_more_neighbours_than_the_default_buffer`
- `test_divergence_and_curl_use_the_periodic_minimum_image` — both operations, against the wrapped displacements from `pair_blocks`
- `test_scatter_add_agrees_across_both_strategies` — both branches, 1-D and 2-D output, against an independent reference
- `test_mass_weighting_reproduces_a_constant_field`
- `test_weighting_is_accepted_and_changes_the_result` — all four operations, and asserts the default is unchanged

The existing C++-versus-python cross-check for `sph_divergence` now runs over both weightings, so the two implementations cannot drift apart.

For scale, with the buffer fix in place `sph_divergence` and `sph_curl` complete over a 42M particle gas snapshot in 14.1 s and 14.6 s.

No version bump: v2.5.0 is the most recent tag, so 2.6.0 is unreleased and this folds into it.

https://claude.ai/code/session_01YGC9utrm2V3Tbu6o6GwgK5

### DIFF
--- a/pynbody/kdtree/__init__.py
+++ b/pynbody/kdtree/__init__.py
@@ -699,7 +699,7 @@ class KDTree:
 
         .. math::
 
-          r_i = \sum_{j=0}^{N_{smooth}}\left( \frac{m_j}{\rho_j} q_i W(|x_i - x_j|, s_i) \right)
+          r_i = \sum_{j=0}^{N_{smooth}}\left( \frac{m_j}{\rho_j} q_j W(|x_i - x_j|, s_i) \right)
 
         where r is the result of the smoothing, m is the mass array, rho is the density array, q is the input array,
         s_i is the smoothing length, and W is the kernel function.
@@ -715,20 +715,27 @@ class KDTree:
             Number of neighbours to use when smoothing.
 
         weighting : str
-            Which average over the kernel to compute. The two agree where the
-            density is nearly constant across a kernel, and differ where it is
-            not.
+            Which average over the kernel to compute.
 
-            * ``'volume'`` (default): the volume-weighted average,
-              :math:`\int f W \, dV`, taking each neighbour's own volume
-              :math:`m_j/\rho_j`. This is the direct discretisation of the
-              smoothing integral.
-            * ``'mass'``: the mass-weighted average,
-              :math:`\int f \rho W \, dV / \int \rho W \, dV`, taking
-              :math:`m_j/\rho_i`. Since :math:`\rho_i` is itself
-              :math:`\sum_j m_j W_{ij}`, the weights sum to one exactly, so
-              this reproduces a constant field exactly and can never fall
-              outside the range of its input.
+            Writing :math:`i` for the particle whose value is being computed
+            and :math:`j` for each of its neighbours, both options evaluate
+            :math:`\sum_j V_j q_j W(|x_i - x_j|, s_i)`, differing only in the
+            volume :math:`V_j` given to each neighbour:
+
+            * ``'volume'`` (default): :math:`V_j = m_j/\rho_j`, each
+              neighbour's own volume. The sum is then the direct
+              discretisation of the smoothing integral, giving the
+              volume-weighted average :math:`\int q W \, dV`.
+            * ``'mass'``: :math:`V_j = m_j/\rho_i`, using the density at the
+              particle being evaluated, which gives the mass-weighted average
+              :math:`\int q \rho W \, dV / \int \rho W \, dV`. Since
+              :math:`\rho_i` is itself :math:`\sum_j m_j W(|x_i - x_j|, s_i)`,
+              the weights here sum to one exactly, so this reproduces a
+              constant field exactly and can never return a value outside the
+              range of its input.
+
+            The two agree where the density is nearly constant across a
+            kernel, and differ where it is not.
 
             For a divergence, ``'mass'`` is the form that satisfies the
             continuity equation exactly, and so is the one appearing in the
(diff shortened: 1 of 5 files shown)


apontzen:
Fixed in 8f228a7. `sph_divergence`'s docstring was copy-pasted from `sph_curl`, so both the parameter and the return description claimed the wrong operation.

Pre-existing on master rather than introduced here, but this PR is already editing those two lines to document `density_weighting`, so it belongs with them rather than in a separate change. `sph_curl` was correct and is untouched.

PR #1019, opened 2026-08-20
Report only ND parents, not their 1D slices, as loadable in CopyOnAccessSimSnap

### LINKED ISSUE #1018: Unable to obtain vz_disp
I have been trying to obtain vertical velocity dispersion from a simulation output. But unfortunately, I am getting the following error:

_KeyError: 'vz_disp is not a valid profile'_

I used the following code:

```
s=pynbody.load('...g2')
s.set_units_system(velocity='km s^-1', mass = '1e10 Msol', distance = '1 kpc') #confirm set_units_system
s['eps'] = pynbody.array.SimArray([1.0],'pc',dtype=np.float32)
pynbody.analysis.center(s, mode='com', with_velocity=False)
h1 = s[sph]
ps = pynbody.analysis.profile.Profile(h1.s,max='15 kpc',min='0 kpc',nbins=20)
print(ps['vz_disp'].in_units('km s^-1'))

```

I though it should be straightforward. Also, even vx_disp, vy_disp are showing similar error.

apontzen:
Fixes #1018 (via #1020, now merged into this branch — see the note at the end).

`CopyOnAccessSimSnap` presents the arrays held by another snapshot as loadable. It built that list by unioning the parent's `loadable_keys()` with its `keys()` — and the latter contains the 1D slices that accompany every ND array, since `_create_array('vel', ndim=3)` also registers `vx`, `vy` and `vz` in `_arrays`.

The copy therefore advertised `vz` as loadable, which no file-backed snapshot ever does — those report the blocks they actually store, i.e. `vel` alone:

```python
f = pynbody.new(10)
f['pos'] = ...; f['vel'] = ...; f['mass'] = ...

f.get_copy_on_access_simsnap().loadable_keys()
# before: ['mass', 'pos', 'vel', 'vx', 'vy', 'vz', 'x', 'y', 'z']
# after:  ['mass', 'pos', 'vel']
```

That divergence matters because code which inspects `loadable_keys()` to work out what a snapshot can provide gets a different answer depending on whether it is looking at a real loader or a copy-on-access view — so a bug that shows up with one is invisible with the other.

## The change

Report just the ND parents. The slices remain fully accessible: a request for one is mapped onto its parent by `_load_array_and_perform_postprocessing`, so `f_c['vz']` still works and pulls in `vel`.

A 1D name is only dropped when its parent is genuinely present in the key set, so a name that merely looks like a slice is left alone. This mirrors the guard already documented in `SimSnap._array_name_implies_ND_slice`.

## Tests

Three regression tests in `copy_on_access_test.py`, all failing on `master` and passing here:

- `loadable_keys()` reports exactly `{pos, vel, mass}` rather than the nine names including slices
- slices still load, and asking for one brings its parent into memory
- a loaded slice is a *view* onto the copy's own ND array rather than a separately stored copy — `np.shares_memory(f_c['vx'], f_c['vel'][:, 0])`, with write-through checked in both directions, and the copy still isolated from the snapshot it copied from

None of them need downloaded test data.

## What this branch now contains

#1020 has been merged into this branch, so the diff here is both changes:

1. this one — `CopyOnAccessSimSnap.loadable_keys` reporting only ND parents
2. **the fix for #1018** — `Profile`/`QuantileProfile` deciding what can be profiled by asking the snapshot rather than looking the name up in `keys`/`all_keys`, so `p['vz_disp']` no longer depends on whether `vel` happens to have been loaded

Because #1020 merged into this branch rather than into `master`, its `Fixes #1018` will not close the issue on its own; the reference at the top of this description covers that when this PR lands.

Both were green across the full 24-check matrix before the merge, and a separate follow-up is filed as #1021.

---

### DIFF
--- a/tests/copy_on_access_test.py
+++ b/tests/copy_on_access_test.py
@@ -1,4 +1,5 @@
 import numpy as np
+import numpy.testing as npt
 import pytest
 
 import pynbody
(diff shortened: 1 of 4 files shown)


tests/copy_on_access_test.py
```diff
@@ -162,3 +163,76 @@ def test_only_try_loading_once():
...
+    f_c['vy'] = np.arange(10, 20)
+    npt.assert_allclose(f_c['vel'][:, 1], np.arange(10, 20))
+
+    # the copy remains isolated from the snapshot it copied from, which is the whole point of
+    # copy-on-access: none of those writes may reach the original
+    assert not np.shares_memory(f_c['vel'], f['vel'])
+    npt.assert_allclose(f['vel'], original_vel)
+
+
+def test_loadable_keys_reports_nd_parent_of_a_generically_named_slice():
```
apontzen: I don't think this test is useful; it's really testing some underlying behaviour regarding making ND arrays from 1D requests which is independent.

apontzen: Agreed — removed in a604096.

You're right that the premise is generic `_create_array` behaviour rather than anything to do with copy-on-access. `f['acc_x'] = np.arange(10)` on the plain in-memory snapshot already gives:

```
base keys: ['acc', 'acc_x', 'acc_y', 'acc_z', 'mass', 'pos', 'vel', 'vx', 'vy', 'vz', 'x', 'y', 'z']
f['acc'].shape: (10, 3)
```

so `acc` exists before `get_copy_on_access_simsnap()` is called at all, and the assertions add nothing over `test_loadable_keys_reports_nd_arrays_not_their_slices`.

Worth recording what I was originally reaching for: the filter is `[k for k in keys if self._array_name_1D_to_ND(k) not in keys]`, and the `not in keys` guard exists so a 1D array with no ND parent is left alone. That situation can't be constructed from an in-memory parent — assigning `acc_x` always creates `acc` — so the test I ended up with asserted more or less the opposite of what I set out to check. The guard only matters if an underlying loader ever advertises a bare `<name>_x` block, so it stays as defensive coding explained by the comment, not something worth a test.

---

PR #1020, opened 2026-08-20
Fix profiles of 1D array slices when the ND array is not yet loaded

apontzen:
Fixes #1018. **Stacked on #1019** — review that one first; this PR targets its branch, so the diff shown here is just the profile change.

## The bug

`Profile._get_profile` decided whether a name referred to a real array by testing it against `sim.keys()` and `sim.all_keys()`. Neither reports the 1D slice names of an ND array unless that array is already in memory:

- `keys()` returns `_arrays.keys()`. `_create_array('vel', ndim=3)` does register `vx`/`vy`/`vz` alongside `vel` — but only once `vel` has been created.
- `loadable_keys()` reports the blocks on disk: `['pos', 'vel', 'mass']`, never `vz`.

So on a freshly loaded snapshot `vz` matched nothing:

```
KeyError: 'vz_disp is not a valid profile'
```

Touch the velocities first and the same call works. Reproduced on a Gadget-2 file matching the reporter's setup:

```
loadable_keys: ['pos', 'vel', 'mass']
'vz' in all_keys: False | 'vz' in keys: False
  vz_disp: FAILED -> 'vz_disp is not a valid profile'
--- with vel preloaded ---
  vz_disp: OK [0.91654175 1.00766647]
```

Two details worth noting:

- **It is format-dependent.** Tipsy stores pos and vel in one record, so loading either loads both and the bug never appears there. The reporter's `.g2` is Gadget-2, which loads blocks independently.
- **It is not just `_disp`.** Plain `p['vz']`, `vz_rms`, `vz_med`, `d_vz` and `QuantileProfile` all failed the same way. The reporter's `p['x']` would have worked, but only incidentally — building the profile computes `rxy`, which loads `pos`.

## The fix

Availability is now established by asking the simulation for the array rather than guessing from its name. `keys()`/`all_keys()` are unreliable in a third way too: a name in `derivable_keys()` is merely *registered*, so the derivation may still fail when it is actually run.

`_get_profile` enumerates the array names a profile name could be built from — the exact name first, then the `_disp`/`_rms`/`_med` interpretations — and tries each. Nothing is wasted: the probed array stays in memory for `_auto_profile`, which was going to fetch it anyway. The `d_` branch likewise just tries `self[name[2:]]`, which recurses through `_get_profile` and so covers registered profiles, arrays and suffixes uniformly, replacing three separate name checks.

Because a failure can no longer be distinguished from "not a profile name" by inspection, the underlying error is attached to the result:

```python
raise KeyError(name + " is not a valid profile") from self._last_array_error
```

This also removes a false positive that a name-based check has: `_array_name_1D_to_ND` guesses `mass_x` is a slice of `mass`, so `p['mass_x']` used to leak `'No array mass_x for family star'` out of a profile lookup. It now correctly reports `'mass_x is not a valid profile'`.

## Trade-off worth flagging

If an array literally named `foo_disp` exists but its derivation raises `KeyError`, the loop now falls back to computing the dispersion of `foo` rather than propagating. That happens only for `KeyError` — the "not available" signal; a derivation failing for a substantive reason (units, bad data) raises something else and propagates as before, and when nothing succeeds the original error is preserved in the chain. Easy to reverse if you would rather the exact-name failure always win.

## Tests

The tests emulate lazy loading with `CopyOnAccessSimSnap`, so they need no downloaded test data — this is what #1019 enables. Each asserts that the lazy and preloaded paths give *identical results*, not merely that nothing raises.

All six `vz` tests fail against `master` with the exact error from the issue, and — importantly for the stack — still fail with #1019 alone:

```
FAIL test_profile_of_1d_slice_without_preloading[vz_disp] -> KeyError 'vz_disp is not a valid profile'
FAIL test_quantile_profile_of_1d_slice_without_preloading -> KeyError 'vz is not a valid QuantileProfile'
```

There is also a test for the registered-but-broken derivation case: a derived array whose dependency does not exist. It is in `all_keys()`, so a name check would call it available; the test asserts the real cause survives into the raised `KeyError`. Removing the `from` clause makes it fail, so the chaining is pinned.

## Verification

Every data-free test in `profile_test.py` (16), `copy_on_access_test.py` (11) and `snapshot_test.py` (18) passes. I could not run the data-backed tests — `api.osf.io` is blocked by this environment's network policy — so CI will be the first full run.

---

### DIFF
--- a/pynbody/analysis/profile.py
+++ b/pynbody/analysis/profile.py
@@ -39,7 +39,8 @@ class Profile:
     can be accessed as ``p['ar']`` where ``p`` is a ``Profile`` object. For example, ``p['vr']`` gives
     the radial velocity profile. Implicitly, this is averaged over all particles in each bin,
     weighted by mass (unless an alternate weighting scheme is passed to the ``weight_by`` keyword argument of the
-    constructor).
+    constructor). Individual components of an ND array may be profiled in the same way, e.g. ``p['vz']`` gives
+    the profile of the z-component of ``vel``.
 
     **Dispersions**: One may append ``_rms`` or ``_disp`` to the name of a
     defined array to get the root-mean-square or dispersion profile, respectively. For example,
@@ -64,6 +65,15 @@ class Profile:
     looking at the same set of particles, centered in the same way. It also means you *must* use the same centering
     method if you want to reuse a saved profile.
 
+    .. versionchanged:: 2.6.0
+
+      What can be profiled is now decided by what the snapshot is actually able to provide, rather than by whether
+      the name appears in ``keys`` or ``all_keys``. Nothing has to be loaded or derived in advance. In particular,
+      a derived array is profiled whenever the snapshot knows how to derive it, and a 1D slice of an ND array is
+      profiled whenever the parent array is available: ``p['vz']`` and ``p['vz_disp']`` now work on a freshly loaded
+      snapshot, where previously they raised ``KeyError`` unless something had already touched the velocities,
+      because a loader offers ``vel`` rather than ``vz``.
+
     .. versionchanged:: 2.0
 
       The method ``create_particle_array`` has been removed. Its behaviour was poorly defined in v1, and not believed
@@ -72,6 +82,10 @@ class Profile:
     """
 
     _profile_registry = {}
+
+    # the failure from the most recent attempt to get an underlying simulation array, used
+    # to explain why a profile could not be constructed
+    _last_array_error = None
 
     def _calculate_x(self, sim):
         if self._x_calculator is not None:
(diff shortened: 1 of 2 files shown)


pynbody/analysis/profile.py
```diff
@@ -64,6 +65,18 @@ class Profile:
     looking at the same set of particles, centered in the same way. It also means you *must* use the same centering
     method if you want to reuse a saved profile.
 
+    .. versionchanged:: 2.6.0
+
+      Profiles of a 1D slice of an ND array, such as ``p['vz']`` or ``p['vz_disp']``, no longer depend on whether
+      the ND array happens to have been loaded already. Previously the underlying array had to be in memory, or
+      named in ``loadable_keys``, for the profile to be recognised; since a loader offers ``vel`` rather than ``vz``,
+      asking for ``p['vz_disp']`` on a freshly loaded snapshot raised ``KeyError``, while the same call succeeded
```
apontzen: For many people I suspect the bigger practical difference is derived arrays will now profile, even if they have not been pre-computed. Certainly deserves a mention here.

pynbody/analysis/profile.py
```diff
@@ -64,6 +65,18 @@ class Profile:
...
+      once anything had touched the velocities.
+
+      Relatedly, whether a profile can be built is now established by asking the snapshot for the underlying array,
```
apontzen: This paragraph is getting too much into the weeds of implementation detail; suggest removing.

pynbody/analysis/profile.py
```diff
@@ -282,41 +299,62 @@ def _get_profile(self, name):
...
+            self._profiles[name] = self._auto_profile(array_name, **kwargs)
             self._profiles[name].sim = self.sim
             return self._profiles[name]
 
-        elif name[-4:] == "_med" and (name[:-4] in list(self.sim.keys()) or name[:-4] in self.sim.all_keys()):
-            logger.info("Auto-deriving %s" % name)
-            self._profiles[name] = self._auto_profile(name[:-4], median=True)
-            self._profiles[name].sim = self.sim
-            return self._profiles[name]
+        if name[0:2] == "d_":
```
apontzen: It now looks very strange that this is handled as a separate case. Surely it should be in `_auto_profile_candidates`, and the derivative mechanism handled by passing a suitable keyword onwards to _auto_profile rather than out here?

pynbody/analysis/profile.py
```diff
@@ -962,6 +1000,12 @@ def _setup_bins(self):
 
 class QuantileProfile(Profile):
     """A profile object that returns requested quantiles instead of means in each bin.
+
+    .. versionchanged:: 2.6.0
+
+      As for :class:`Profile`, quantiles of a 1D slice of an ND array, such as ``p['vz']``, no longer depend on
```
apontzen: As with the Profile case, ensure this notes derived arrays just as much as named slices of arrays

tests/profile_test.py
```diff
@@ -200,3 +200,81 @@ def test_quantile_profile(weight):
         # for where the cdf is 0.16 and 0.84
         expected_width *= 1.8724
     npt.assert_allclose(np.diff(pro['testquantity'], axis=1), expected_width, atol=2.5e-2)
+
+
+def _make_lazy_loading_snapshot(npart=1000):
```
apontzen, line 205: would this be better as a fixture (`lazy_loading_snapshot`) that tests can take?

pynbody/analysis/profile.py
apontzen: Agreed, done in abc3520. `_auto_profile_candidates` now yields `(name[2:], {'derivative': True})`, and `_auto_profile` grew a `derivative` keyword that returns early with the gradient. `_get_profile` is down to the candidate loop plus the final raise.

The one wrinkle is that a derivative differentiates *another profile* rather than averaging a snapshot array, so the availability check had to become source-aware. `_sim_array_obtainable` is now `_source_obtainable`, which looks the source up with `self[...]` for a derivative and `self.sim[...]` otherwise.

Folding it in fixes a case that fell through the old branch. Its guard was `name[2:] in self.keys() or ... or name[2:] in self.sim.all_keys()`, so `d_vz_disp` never qualified — `vz_disp` is neither a profile key nor a snapshot array:

```
                master        this PR
d_vz_disp       KeyError      OK
```

Everything else is unchanged; I diffed the derivative profiles against master and the values are identical:

```
density        OK  Msol kpc**-2         [281.38593939 230.03194442]
d_density      OK  Msol kpc**-3         [-102.70798994 -138.27381456]
d_mass_enc     OK  Msol kpc**-1         [1084. 1104.]
d_vz           OK  km s**-1 kpc**-1     [-0.25220176 -0.04686979]
d_nonsense     KeyError
```

`d_vz`'s units print as `km s**-1 kpc**-1` where master printed `km kpc**-1 s**-1` — same unit, different ordering in the string (`Unit('km kpc**-1 s**-1') == Unit('km s**-1 kpc**-1')` is True, ratio 1).

---

pynbody/analysis/profile.py
apontzen: Mentioned in abc3520 — but I want to flag a nuance, because I checked before writing it and it isn't quite the whole picture.

A *registered* derived array was already in `all_keys()` (via `derivable_keys()`), so on an ordinary snapshot it already profiled without being pre-computed. Comparing master against this branch with a derived array that has never been touched:

```
                     all_keys  derivable   master              this PR
plain                True      True        OK                  OK
family subsnap .s    True      True        OK                  OK
indexed subsnap      True      True        OK                  OK
copy-on-access       False     True        KeyError            OK
coa family .s        False     True        KeyError            OK
```

So the case that genuinely changes is where `all_keys()` under-reports what the snapshot can derive. `UnderlyingClassMixin` overrides `find_deriving_function` to consult `_underlying_class`, but `derivable_keys()` still only walks `type(self).__mro__` — so for a `CopyOnAccessSimSnap` emulating another class the derivation is available while the name is absent from `all_keys()`. That's the tangos shared-memory path, so "many people" may be generous, but it was a real hole.

I've written the note around what's now true in general rather than claiming derived arrays didn't work before:

> What can be profiled is now decided by what the snapshot is actually able to provide, rather than by whether the name appears in ``keys`` or ``all_keys``. Nothing has to be loaded or derived in advance. In particular, a derived array is profiled whenever the snapshot knows how to derive it, and a 1D slice of an ND array is profiled whenever the parent array is available […]

Happy to lead with the derived-array case if you'd rather it came first — say the word. Separately, the `derivable_keys()`/`find_deriving_function` mismatch in `UnderlyingClassMixin` looks like a bug in its own right; I've left it alone as out of scope here, but can open an issue.

---

pynbody/analysis/profile.py
apontzen: Removed in abc3520. The `keys`/`all_keys` mechanics and the exception chaining are gone from the class docstring; the rationale now lives only in the `_source_obtainable` docstring where it belongs.

---

pynbody/analysis/profile.py
apontzen: Done in abc3520 — `QuantileProfile` now reads:

> As for :class:`Profile`, what can be profiled is decided by what the snapshot is able to provide rather than by the names it advertises, so derived arrays and 1D slices of ND arrays are quantiled without having to be computed or loaded in advance.

Same caveat as on the `Profile` thread about which derived-array cases actually changed.

`QuantileProfile._get_profile` also needed updating for the rename to `_source_obtainable`.

---

tests/profile_test.py
apontzen: Yes — done in abc3520. Two fixtures over a shared builder, since the comparison tests need two independent snapshots:

```python
@pytest.fixture
def lazy_loading_snapshot():
    """A snapshot which lazily copies its arrays from another, as a file-backed one would. …"""
    return _make_lazy_loading_snapshot()


@pytest.fixture
def preloaded_snapshot():
    """The same snapshot, but with the velocities already pulled into memory."""
    f = _make_lazy_loading_snapshot()
    f['vel']
    return f
```

so the tests read `def test_profile_of_1d_slice_without_preloading(name, lazy_loading_snapshot, preloaded_snapshot)` and lose their setup boilerplate.

Worth noting this caught a gap in how I'd been checking these locally: `api.osf.io` is blocked in my sandbox, so the module-scoped `get_data` fixture stops me running `profile_test.py` under pytest at all, and I'd been calling the test functions directly instead. That silently skips fixture-based tests. I've since been running the new tests through real pytest from an extracted copy, with `pyproject.toml`'s `filterwarnings` mirrored so warnings are errors — 8 passed.

---

pynbody/analysis/profile.py
apontzen: Good points, do open a separate issue regarding this mismatch

apontzen: Opened as #1021, with the reproducer and a suggested fix (override `derivable_keys` in `UnderlyingClassMixin` to mirror `find_deriving_function`).

Leaving the docstring wording as it stands, since you were happy with the framing.

---

PR #1025, opened 2026-09-01
Fix incorrect time written by TipsySnap._write for non-cosmological snapshots

### LINKED ISSUE #996: Incorrect time output from tipsy._write()
**Describe the bug**
When writing a new Tipsy snapshot, the time stored in the output file is not the same as the time associated with the snapshot being written.

**To Reproduce**
```python
import pynbody
snap = pynbody.load('snapshot_filepath')
snap.write(fmt=pynbody.snapshot.tipsy.TipsySnap, filename='NewSnap.bin', binary_aux_arrays=True, cosmological=False)
``` 
```python
new_snap = pynbody.load('NewSnap.bin')
new_snap.properties['time'].in_units('Gyr')
``` 
```python
snap.properties['time'].in_units('Gyr')
``` 

Expected the output snapshot file to have the same time as the input snapshot.

 - OS: Linux 6.14
 - Python version: 3.12.3
 - Numpy version: 2.3.3
 - Pynbody version: 2.3.1

apontzen:
## Summary

Fixes #996, where writing a non-cosmological Tipsy snapshot back out (`cosmological=False`) stores the wrong time in the header.

**Root cause**: On load, `TipsySnap.properties['time']` is set to the raw header value multiplied by a paramfile-inferred time unit (`self.properties['time'] *= time_unit` in `TipsySnap.__init__`). Because `Unit * float` folds the raw value directly into a `Unit`'s scale, `float(properties['time'])` no longer returns anything recoverable as the original raw header value — it returns an intermediate, not-directly-meaningful number. `TipsySnap._write` packed this straight into the header via `struct.pack`, which implicitly calls `float()` on it, so every write baked in an extra, spurious conversion factor (and this compounded further on repeated round trips).

**Fix (revised after discussion)**: rather than special-casing `_write` to convert the value back (an earlier version of this PR did that), `properties['time']` is now represented as a `SimArray` — the raw, file-native value with the time unit attached as a label, exactly like any other time-valued array in pynbody (e.g. `star['tform']`). `float()`/plain arithmetic on it recovers the raw value (so `_write` needs no special-casing at all — it's back to its original, two-line form), while `.in_units(...)` still gives the physical value on request. This is also more consistent for users: previously `properties['time']` behaved quite differently from other time-carrying quantities (e.g. it wouldn't interoperate directly with `star['tform']` the way two arrays would); as a `SimArray` it now does.

`physical_units()`/`set_units_system()` already have working, tested code paths for `SimArray`-typed properties, so this required no changes elsewhere — confirmed the existing test suite (tipsy, partial-tipsy, derived arrays, snapshot, nchilada, gadget, ramses, swift, pkdgravhdf) all still pass, and that derived quantities like `star['age']` produce bit-identical results to before.

## Test plan

- [x] `test_time_roundtrip_noncosmological` in `tests/tipsy_test.py`: builds a synthetic non-cosmological Tipsy snapshot (with its own `.param` file), checks `properties['time']` is a `SimArray` equal to the raw header value, and that a write → load → write → load round trip preserves both the raw header value and the physical time in Gyr. Confirmed this test fails against the original code (reproducing the exact corruption) and passes with the fix.
- [x] Ran `tests/tipsy_test.py` + `tests/partial_tipsy_test.py` (38 tests), plus `tests/derived_arrays_test.py`, `tests/snapshot_test.py`, `tests/nchilada_test.py`, `tests/gadget_test.py`, `tests/ramses_test.py`, `tests/swift_test.py`, `tests/pkdgravhdf_test.py` — all pass.
- [x] Verified `star['age']` (a derived array depending on `properties['time']`) and `properties['time'].in_units('Gyr')` produce identical numeric results before and after this change on a real cosmological test snapshot.

https://claude.ai/code/session_016JnfB7JNzuWcJQxj7T7df1

### DIFF
--- a/pynbody/snapshot/tipsy.py
+++ b/pynbody/snapshot/tipsy.py
@@ -135,7 +135,13 @@ class TipsySnap(SimSnap):
             pass
 
         if time_unit is not None:
-            self.properties['time'] *= time_unit
+            # Store as a SimArray (raw value + unit label) rather than folding the
+            # conversion into a bare Unit, consistent with how every other time-valued
+            # quantity (e.g. star['tform']) is represented. This keeps the raw,
+            # file-native value recoverable (needed so _write can round-trip it) while
+            # still allowing sim.properties['time'].in_units(...) to give the physical
+            # value, and lets it interoperate directly with other time SimArrays.
+            self.properties['time'] = array.SimArray(self.properties['time'], time_unit)
 
     def _load_main_file(self):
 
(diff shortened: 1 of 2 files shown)


apontzen:
Human note for future reference. Storing the time as a SimArray singleton is awkward, but if it is stored as a Unit then there is no logical way to disambiguate the user's intent given the accepted behaviour that arrays write in their current units (i.e., they don't automatically convert back to paramfile-inferred units -- to change that now would be a severe breaking change).

PR #1026, opened 2026-09-02
Allocate halo catalogue index lists in shared memory when the snapshot asks for it

apontzen:
## The problem

A halo catalogue's particle index list is $8\,\mathrm{bytes} \times N_\mathrm{particles\ in\ halos}$, i.e. multiple GB for a large simulation. A consumer that loads a catalogue once and shares it with worker processes — [tangos](https://github.com/pynbody/tangos) being the motivating case — had no way to avoid copying the whole thing.

Instrumenting the 2.7.0 API over tangos's full test suite: 22 of 22 calls copy, and not one array of a portable state is ever already in shared memory. Two things caused that:

1. **The index lists were allocated with plain numpy.** `HaloCatalogue._assemble_particle_indices`, and the independent implementations in `number_array`, `subfind`, `subfindhdf` and `rockstar`, used `np.empty`/`np.zeros` regardless of what the parent snapshot wanted.
2. **`get_portable_state` laundered the arrays through `np.asarray`.** That returns a base-class view, which has lost `_shared_fname` and so cannot be turned into a `SharedArrayReference` at all — even if the array had been allocated shared.

### Allocation

`SimSnap` gains a public `uses_shared_arrays()` accessor. The underlying `_shared_arrays` flag is per-`SimSnap` and is *not* inherited by subsnaps, so reading it off a view gave the wrong answer; the accessor resolves the ancestor. `SimSnap._create_array`/`_create_family_array` now use it too, so there is one source of truth, and neither `halo/` nor external code (tangos sets the flag directly today) needs to touch the private attribute.

`HaloCatalogue` gains two helpers built on it:

- `_array_factory(dims, dtype, zeros)` — allocates via `pynbody.array.array_factory`, following the `SimSnap._create_array` precedent. No copy.
- `_adopt_array(source)` — for arrays numpy has already allocated on our behalf (`argsort`, `concatenate`), where there is no way to allocate into a shared buffer up front. Returns `source` unchanged when shared arrays are off or the array is already shared; otherwise copies it once. Every use site says explicitly that this path copies.

`_assemble_particle_indices` allocates through `_array_factory`, which covers the ahf, subfind, rockstar, hbtplus, velociraptor and adaptahop catalogues. `number_array` (the `grp` catalogue), `subfind` and `subfindhdf` are converted individually.

Not converted: `SubFindHDFHaloCatalogue._get_particle_indices_one_halo`. That is the single-halo path, and giving each accessed halo its own shared segment would be worse than the copy it avoids.

### Portable state

The encoding sites now use a new `as_portable_array` helper instead of `np.asarray`. It is `np.asanyarray` plus a guard: pynbody's own array types pass through with their identity intact, anything else (a masked array, `np.matrix`) is still reduced to a plain `ndarray`, so a recipient can still assume ordinary semantics. `SimArray` with units is fine because the property encoder records units separately.

Finder property arrays are handled too, "where practical". `properties_to_portable_state` takes an `adopt` callable which `HaloCatalogue.get_portable_state` supplies from `_adopt_array`. This is not a flag on the public `get_portable_state` signature — the catalogue derives it from the snapshot — and it is needed because some of the arrays in a properties state are *derived at encode time* (the concatenation of a ragged property such as `children`), so there is no earlier point at which they could have been allocated.

Result: with `f.enable_shared_arrays()` before `f.halos()`, every array of the state is in shared memory. Verified across ahf, gadget-4 subfind, subfind HDF (gadget-3), plain subfind, HBT+, rockstar and adaptahop against the test data.

### Ownership — a contract change

`SharedMemorySimArray.__del__` unlinks the segment when the owning array is collected. Previously a consumer made its own shared copies and the originating catalogue was free to die. Now the originating process owns the memory, so:

> **Keep the state itself alive for as long as any other process may still be using the arrays.** Holding the catalogue is not enough (a few property arrays are derived by, and held only by, the state), and neither is holding the references, which describe the memory without keeping it allocated.

This is spelled out in `get_portable_state`, `from_portable_state`, `HaloParticleIndices.get_portable_state`, the `pynbody.halo.portable` module docstring, and the parallelism tutorial.

On `transfer_ownership`: it does not give consumers a clean way to hand ownership over, because `from_shared_reference` has never supported reconstructing an owning reference. Rather than leave that as a bare `assert`, it now raises `NotImplementedError` with an explanation, and the `to_shared_reference` docstring carries a warning.

### `to_shared_reference` on a plain ndarray

Previously this fell through to `_recursive_shared_array_deconstruct`'s `for item in input:` loop and iterated the array:

```python
to_shared_reference(np.asarray(make_shared_array((6,), np.int64, False)))
# -> [np.int64(0), np.int64(1), ..., np.int64(5)]
```

So a caller passing a large plain ndarray got a python list of every element, silently. It now raises `TypeError`, as does a `SimArray` that is simply not in shared memory. This matters more after the change above, since callers will start expecting state arrays to be referenceable. A new `is_shared_array()` predicate lets them check first. The nested-item behaviour inside `remote_map` is deliberately unchanged — ramses parallel reading relies on it.

### Large-scale copies removed from halo internals

- `HaloNumberCatalogue` derived its halo boundaries from `np.searchsorted` against a **full sorted copy of the whole per-particle array**. `np.unique(..., return_counts=True)` gives the counts from the pass it was already making, so the boundaries come from a cumsum with no $O(N_\mathrm{part})$ temporary and no `hstack`.
- `RockstarCatalogue._get_all_particle_indices` and `SubfindCatalogue._read_ids` grew their results with `np.append` in a loop, copying everything read so far once per finder output file — quadratic in the number of files. Both accumulate and concatenate once.

## Testing

New tests in `tests/halos_test.py` and `tests/array_test.py` cover: the accessor resolving through subsnaps; index lists (and boundaries, and missing-particle counts) being `SharedMemorySimArray` after `load_all()` for the `grp` catalogue, the assembling path, and AHF on real test data; every array of a state being referenceable with no intermediate copy, and identical to the array the catalogue holds; a same-process round trip through `from_portable_state`; an end-to-end transfer to a spawned subprocess with no copying pass at all; nothing changing when shared arrays are off; and `to_shared_reference` raising `TypeError`. The reference implementation in `halos_test.py` was updated to use the new helpers, so it demonstrates the pattern a real subclass should follow.

Full suite: 1028 passed, 1 skipped, 1 failed — `ahf_halos_test.py::test_ahf_unwritable`, which fails identically on `master` in this environment because the tests run as root and `chmod` therefore does not make the directory unwritable.

Version bumped 2.7.0 → 2.7.1.

## Sequencing with tangos

pynbody/tangos#302 contains `_as_shared_memory_array`, which does the copying described above and is unconditional precisely because nothing ever arrives shared. Once this lands, that becomes a cheap "already shared? then just reference it" check (`is_shared_array` is there for exactly this), and tangos can drop the copy. tangos also currently sets `current_snapshot._shared_arrays = True` directly; `uses_shared_arrays()` lets both sides stop touching the private flag. No tangos change is needed for this PR to be correct — the copy simply becomes redundant.

https://claude.ai/code/session_01J5AL212J957VERirvoP5FX

---
_Generated by [Claude Code](https://claude.ai/code/session_01J5AL212J957VERirvoP5FX)_


apontzen:
The `docs/readthedocs.org:pynbody` check is red here, but it is not this PR's failure — it is red on `master` too, and has been since before this branch existed.

The build dies in the `pre_build` step from `.readthedocs.yaml`, which downloads the extended test data from OSF:

```
cd tests && python -c "import pynbody.test_utils as tu; tu.precache_test_data(extended=True)"
...
  File ".../osfclient/models/file.py", line 64, in write_to
    raise RuntimeError("Response has status code {}.".format(response.status_code))
RuntimeError: Response has status code 403.
```

It gets ~1.16 GB in and then OSF returns a 403. Read the Docs build history for the project:

| build | version | commit | result |
|---|---|---|---|
| 34362784 | this PR | `fdfef09` | failed — OSF 403 |
| 34349202 / 34349200 | `v2.7.0` | `f8ade17` | failed |
| 34349201 | `stable` | `f8ade17` | failed — OSF 403 |
| 34348584 | `latest` (master) | `f8ade17` | failed — OSF 500 |
| 34347733 | `latest` (master) | — | last success, 07:28 UTC |

So every docs build since roughly 08:00 UTC today has failed the same way, on `master` and on the release tag as well as here. The diff in this PR touches `pynbody/halo/`, `pynbody/array/shared/`, `pynbody/snapshot/simsnap.py` and `docs/tutorials/parallelism.rst` — nothing in `pynbody/test_utils/`, nothing in the OSF download path, and nothing in `.readthedocs.yaml`.

There is no fix to port: this is OSF availability or rate-limiting, not a change in the repository, so it should clear on its own once OSF is serving those files again. I have no means to trigger a Read the Docs rebuild from here (no RTD credentials), so I can't spend the one re-run on it.

`pre-commit.ci` is green. I'll keep watching this PR and report on `Build and Test`, which is still running, when it finishes.

---

apontzen:
The complexity of this does not justify the theoretical memory overhead savings. Closing for now.

